PR #13, opened 2017-09-03
Remove an unnecessary percentile character in libraryDependencies on README.md

richardimaoka:


adelbertc:
I think this is done for Scala.JS's benefit? https://www.scala-js.org/doc/project/dependencies.html

richardimaoka:
Ah ok if it was intentional, let me close this pull request.

Issue #32, opened 2018-06-24
Add a MonadChronicle type class


LukaJCB:
`MonadChronicle` is the mtl equivalent of `IorT` and has some fairly nice properties if you ask me. We should consider adding support for it :)

YuvalItzchakov:
I'd love to take this up if it's still available :)

LukaJCB:
Closed by #35 :)

PR #35, opened 2018-07-20
MonadChronicle implementation

YuvalItzchakov:
This is a WIP of https://github.com/typelevel/cats-mtl/issues/32

Adds:

1. MonadChronicle definition
2. MonadChronicle instance for IorT
3. IorT FunctorLayer definition
4. IorT MonadControlLayer definition
5. Basic tests

I would love help with figuring out what the laws for this typeclass should be so I can expand the test suite and add tests for MC. Once we figure that out and see that it holds, I'll add more instance definitions for other MT.

### DIFF
--- a/core/src/main/scala/cats/mtl/MonadChronicle.scala
+++ b/core/src/main/scala/cats/mtl/MonadChronicle.scala
@@ -1,0 +1,29 @@
+package cats
+package mtl
+
+import cats.data.Ior
+
+trait MonadChronicle[F[_], E] extends Serializable {
+  val monad: Monad[F]
+
+  def dictate(c: E): F[Unit]
+
+  def disclose[A](c: E)(implicit M: Monoid[A]): F[A]
+
+  def confess[A](c: E): F[A]
+
+  def memento[A](ma: F[A]): F[Either[E, A]]
+
+  def absolve[A](a: A, ma: F[A]): F[A]
+
+  def condemn[A](ma: F[A]): F[A]
+
+  def retcon[A](cc: E => E, ma: F[A]): F[A]
+
+  def chronicle[A](ior: E Ior A): F[A]
+}
+
+object MonadChronicle {
+  def apply[F[_], E](implicit monadChronicle: MonadChronicle[F, E]): MonadChronicle[F, E] =
+    monadChronicle
+}
--- a/core/src/main/scala/cats/mtl/MonadChronicle.scala
+++ b/core/src/main/scala/cats/mtl/MonadChronicle.scala
@@ -36,17 +36,6 @@ object MonadChronicle {
   def materialize[F[_], E, A](fa: F[A])(implicit ev: MonadChronicle[F, E]): F[E Ior A] =
     ev.materialize(fa)
 
-  def memento[F[_], E, A](fa: F[A])(implicit ev: MonadChronicle[F, E]): F[Either[E, A]] =
-    ev.memento(fa)
-
-  def absolve[F[_], E, A](a: => A, fa: F[A])(implicit ev: MonadChronicle[F, E]): F[A] =
-    ev.absolve(a, fa)
-
-  def condemn[F[_], E, A](fa: F[A])(implicit ev: MonadChronicle[F, E]): F[A] = ev.condemn(fa)
-
-  def retcon[F[_], E, A](cc: E => E, fa: F[A])(implicit ev: MonadChronicle[F, E]): F[A] =
-    ev.retcon(cc, fa)
-
   def chronicle[F[_], E, A](ior: E Ior A)(implicit ev: MonadChronicle[F, E]): F[A] =
     ev.chronicle(ior)
 
@@ -58,8 +47,8 @@ trait DefaultMonadChronicle[F[_], E] extends MonadChronicle[F, E] {
   override def disclose[A](c: E)(implicit M: Monoid[A]): F[A] = monad.as(dictate(c), M.empty)
 
   override def memento[A](fa: F[A]): F[Either[E, A]] = monad.flatMap(materialize(fa)) {
-    case Ior.Left(e) => monad.pure(Left(e))
-    case Ior.Right(a) => monad.pure(Right(a))
+    case Ior.Left(e)    => monad.pure(Left(e))
+    case Ior.Right(a)   => monad.pure(Right(a))
     case Ior.Both(e, a) => monad.as(dictate(e), Right(a))
   }
 
--- a/core/src/main/scala/cats/mtl/MonadChronicle.scala
+++ b/core/src/main/scala/cats/mtl/MonadChronicle.scala
@@ -12,18 +12,74 @@ trait MonadChronicle[F[_], E] extends Serializable {
 
   def confess[A](c: E): F[A]
 
-  def memento[A](ma: F[A]): F[Either[E, A]]
+  def materialize[A](fa: F[A]): F[E Ior A]
 
-  def absolve[A](a: A, ma: F[A]): F[A]
+  def memento[A](fa: F[A]): F[Either[E, A]]
 
-  def condemn[A](ma: F[A]): F[A]
+  def absolve[A](a: => A, fa: F[A]): F[A]
 
-  def retcon[A](cc: E => E, ma: F[A]): F[A]
+  def condemn[A](fa: F[A]): F[A]
+
+  def retcon[A](cc: E => E, fa: F[A]): F[A]
 
   def chronicle[A](ior: E Ior A): F[A]
 }
 
 object MonadChronicle {
-  def apply[F[_], E](implicit monadChronicle: MonadChronicle[F, E]): MonadChronicle[F, E] =
-    monadChronicle
+  def dictate[F[_], E](e: E)(implicit ev: MonadChronicle[F, E]): F[Unit] = ev.dictate(e)
+
+  def disclose[F[_], A, E](c: E)(implicit ev: MonadChronicle[F, E], m: Monoid[A]): F[A] =
+    ev.disclose(c)
+
+  def confess[F[_], E, A](c: E)(implicit ev: MonadChronicle[F, E]): F[A] = ev.confess(c)
+
+  def materialize[F[_], E, A](fa: F[A])(implicit ev: MonadChronicle[F, E]): F[E Ior A] =
+    ev.materialize(fa)
+
+  def memento[F[_], E, A](fa: F[A])(implicit ev: MonadChronicle[F, E]): F[Either[E, A]] =
+    ev.memento(fa)
+
+  def absolve[F[_], E, A](a: => A, fa: F[A])(implicit ev: MonadChronicle[F, E]): F[A] =
+    ev.absolve(a, fa)
+
+  def condemn[F[_], E, A](fa: F[A])(implicit ev: MonadChronicle[F, E]): F[A] = ev.condemn(fa)
+
+  def retcon[F[_], E, A](cc: E => E, fa: F[A])(implicit ev: MonadChronicle[F, E]): F[A] =
+    ev.retcon(cc, fa)
+
+  def chronicle[F[_], E, A](ior: E Ior A)(implicit ev: MonadChronicle[F, E]): F[A] =
+    ev.chronicle(ior)
+
+  def apply[F[_], E](implicit ev: MonadChronicle[F, E]): MonadChronicle[F, E] =
+    ev
+}
+
+trait DefaultMonadChronicle[F[_], E] extends MonadChronicle[F, E] {
+  override def disclose[A](c: E)(implicit M: Monoid[A]): F[A] = monad.as(dictate(c), M.empty)
+
+  override def memento[A](fa: F[A]): F[Either[E, A]] = monad.map(materialize(fa))(_.toEither)
+
+  override def absolve[A](a: => A, fa: F[A]): F[A] = monad.map(materialize(fa)) {
+    case Ior.Left(_)     => a
+    case Ior.Right(a0)   => a0
+    case Ior.Both(_, a0) => a0
+  }
+
+  override def condemn[A](fa: F[A]): F[A] = monad.flatMap(materialize(fa)) {
+    case Ior.Left(e)    => confess(e)
+    case Ior.Right(a)   => monad.pure(a)
+    case Ior.Both(e, _) => confess(e)
+  }
+
+  override def retcon[A](cc: E => E, fa: F[A]): F[A] = monad.flatMap(materialize(fa)) {
+    case Ior.Left(e)    => confess(cc(e))
+    case Ior.Right(a)   => monad.pure(a)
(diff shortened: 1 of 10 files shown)


core/src/main/scala/cats/mtl/MonadChronicle.scala
```diff
@@ -0,0 +1,29 @@
...
+package mtl
+
+import cats.data.Ior
+
+trait MonadChronicle[F[_], E] extends Serializable {
+  val monad: Monad[F]
+
+  def dictate(c: E): F[Unit]
+
+  def disclose[A](c: E)(implicit M: Monoid[A]): F[A]
```
LukaJCB: We can implement this in terms of dictate :)
```scala
def disclose[A](c: E)(implicit M: Monoid[A]): F[A] =
  dictate(c).as(M.empty)
```

core/src/main/scala/cats/mtl/MonadChronicle.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  def memento[A](ma: F[A]): F[Either[E, A]]
+
+  def absolve[A](a: A, ma: F[A]): F[A]
+
+  def condemn[A](ma: F[A]): F[A]
+
+  def retcon[A](cc: E => E, ma: F[A]): F[A]
+
+  def chronicle[A](ior: E Ior A): F[A]
```
LukaJCB: Can we express this in terms of `confess`, `pure` and `dictate`:

```scala
def chronicle[A](ior: E Ior A): F[A] = ior match {
  case Left(e) => confess(e)
  case Right(a) => monad.pure(a)
  case Both(e, a) => dictate(e).as(a)
}
```

core/src/main/scala/cats/mtl/MonadChronicle.scala
```diff
@@ -0,0 +1,29 @@
...
+trait MonadChronicle[F[_], E] extends Serializable {
+  val monad: Monad[F]
+
+  def dictate(c: E): F[Unit]
+
+  def disclose[A](c: E)(implicit M: Monoid[A]): F[A]
+
+  def confess[A](c: E): F[A]
+
+  def memento[A](ma: F[A]): F[Either[E, A]]
```
LukaJCB: After looking at this for a long time, I think we could radically make this type class simpler by returning `F[Ior[E, A]]` here instead. WDYT?

core/src/main/scala/cats/mtl/MonadChronicle.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  def absolve[A](a: A, ma: F[A]): F[A]
```
LukaJCB: Let's take that `A` by-name :)

With memento returning `Ior` this could be:
```scala
def absolve[A](default: => A, ma: F[A]): F[A] =
  memento(ma).map {
    case Left(_) => default
    case Right(a) => a
    case Both(_, a) => a
  }
```

core/src/main/scala/cats/mtl/MonadChronicle.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  def condemn[A](ma: F[A]): F[A]
```
LukaJCB: With memento returning Ior this could be:

```scala
def condemn[A](ma: F[A]): F[A] =
  memento(ma).flatMap {
    case Left(e) => confess(e)
    case Right(a) => pure(a)
    case Both(_, a) => confess(e)
  }
```

core/src/main/scala/cats/mtl/MonadChronicle.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  def retcon[A](cc: E => E, ma: F[A]): F[A]
```
LukaJCB: With memento returning Ior this could be:

```scala
def retcon[A](cc: E => E, ma: F[A]): F[A] =
  memento(ma).map {
    case Left(e) => confess(cc(e))
    case Right(a) => pure(a)
    case Both(e, a) => dictate(cc(e)).as(a)
  }
```

core/src/main/scala/cats/mtl/MonadChronicle.scala
LukaJCB: /cc @SystemFw @andyscott

core/src/main/scala/cats/mtl/instances/iort.scala
```diff
@@ -0,0 +1,100 @@
...
+      override def layerControl[B](
+          cps: (IorTC[M, A]#l ~> (M of IorC[A]#l)#l) => M[B]): IorT[M, A, B] = {
+        IorT[M, A, B] {
+          cps(new (IorTC[M, A]#l ~> (M of IorC[A]#l)#l) {
+            def apply[X](fa: IorT[M, A, X]): M[Ior[A, X]] = fa.value
+          }).map(b => Ior.right[A, B](b))
+        }
+      }
+
+      override def zero[B](state: Ior[A, B]): Boolean = false
```
LukaJCB: This should return `true` if the `Ior` short circuits. I.e. `state.isLeft` :)

core/src/main/scala/cats/mtl/MonadChronicle.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  def chronicle[A](ior: E Ior A): F[A]
+}
+
+object MonadChronicle {
+  def apply[F[_], E](implicit monadChronicle: MonadChronicle[F, E]): MonadChronicle[F, E] =
+    monadChronicle
+}
```
LukaJCB: It'd also be cool to add a `trait DefaultMonadChronicle[F[_]] extends MonadChronicle[F]` :)

core/src/main/scala/cats/mtl/MonadChronicle.scala
YuvalItzchakov: Definitely.

core/src/main/scala/cats/mtl/MonadChronicle.scala
YuvalItzchakov: 👍

core/src/main/scala/cats/mtl/MonadChronicle.scala
YuvalItzchakov: I'm going to go ahead and implement this and see how it plays out with deriving the other combinators while we wait for more input :)

core/src/main/scala/cats/mtl/MonadChronicle.scala
YuvalItzchakov: Since `memento` (or now `materialize`) returns an `F[E Ior A]`, we need to get under the `F`:

```scala
def absolve[A](a: => A, fa: F[A]): F[A] = monad.map(materialize(fa)) {
    case Ior.Left(_) => a
    case Ior.Right(a0) => a0
    case Ior.Both(_, a0) => a0
 }
```

core/src/main/scala/cats/mtl/MonadChronicle.scala
YuvalItzchakov: Yes! On the `Both` case we need `case Ior.Both(e, _) => confess(e)` :)

andyscott:
This looks great! In addition to working on tests, can we also add documentation to the new methods? The names and signatures are mostly intuitive but I think brief descriptions of use would go a long way.

YuvalItzchakov:
Definitely. I just wanted to make sure to API is agreed upon before I proceed. I'll also provide more instances of `MonadChronicle` for additional transformers.

LukaJCB:
Okay, so here's a few things that I think could be the next steps:

1. Add some Scaladoc
2. Add an instance for plain `Ior`
3. Find laws (For this it might make sense to add a `Semigroup[E]` constraint to better define them)

YuvalItzchakov:
Thanks Luka. I hope I can complete this over the weekend.

kailuowang:
hmm somehow the travis build wasn't triggered properly, let me close and reopen.

YuvalItzchakov:
There is one problem I'm facing regarding the implementation of `materialize` and the `condemn(dictate(c)) <-> confess(c)` law. 

Current implementation of `materialize` mimics `memento`, but since we're using `Ior`, we're returning `Ior.both(e, Ior.right(a)` for the `both` case:

```scala
override def materialize[A](fa: IorT[F, E, A]): IorT[F, E, E Ior A] = IorT[F, E, E Ior A] {
        fa.value.map {
          case Ior.Left(e)    => Ior.right(Ior.left(e))
          case Ior.Right(a)   => Ior.right(Ior.right(a))
          case Ior.Both(e, a) => Ior.both(e, Ior.right(a))
        }
      }
```

This breaks the law:

```
[info] - IorT[Option, String, ?].monadChronicle.dictate then condem is confess *** FAILED *** (61 milliseconds)
[info]   GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation.
[info]    (Discipline.scala:14)
[info]     Falsified after 0 successful property evaluations.
[info]     Location: (Discipline.scala:14)
[info]     Occurred when passed generated values (
[info]       arg0 = ""
[info]     )
[info]     Label of failing property:
[info]       Expected: IorT(Some(Left()))
[info]   Received: IorT(Some(Both(,())))
```

WDYT?

core/src/main/scala/cats/mtl/MonadChronicle.scala
```diff
@@ -0,0 +1,85 @@
...
+    ev.chronicle(ior)
+
+  def apply[F[_], E](implicit ev: MonadChronicle[F, E]): MonadChronicle[F, E] =
+    ev
+}
+
+trait DefaultMonadChronicle[F[_], E] extends MonadChronicle[F, E] {
+  override def disclose[A](c: E)(implicit M: Monoid[A]): F[A] = monad.as(dictate(c), M.empty)
+
+  override def memento[A](fa: F[A]): F[Either[E, A]] = monad.map(materialize(fa))(_.toEither)
```
LukaJCB: I think this should be 
```scala
monad.flatMap(materialize(fa)) {
  case Left(e) => pure(Either.left(e))
  case Right(a) => pure(Either.right(a))
  case Both(e, a) => dictate(e).as(Either.right(a))
}
```

Essentially, a `MonadChronicle` should keep its current record unless it ended with `confess`. See the Haskell docs:

> memento m is an action that executes the action m, returning either its record if it ended with confess, or its final value otherwise, with any record added to the current record.

http://hackage.haskell.org/package/these-0.7.4/docs/Control-Monad-Chronicle.html

YuvalItzchakov: Hmm, I see. In the `Both` case we'd truncate the left side. Ok, let me fix that

core/src/main/scala/cats/mtl/MonadChronicle.scala
```diff
@@ -0,0 +1,89 @@
...
+
+object MonadChronicle {
+  def dictate[F[_], E](e: E)(implicit ev: MonadChronicle[F, E]): F[Unit] = ev.dictate(e)
+
+  def disclose[F[_], A, E](c: E)(implicit ev: MonadChronicle[F, E], m: Monoid[A]): F[A] =
+    ev.disclose(c)
+
+  def confess[F[_], E, A](c: E)(implicit ev: MonadChronicle[F, E]): F[A] = ev.confess(c)
+
+  def materialize[F[_], E, A](fa: F[A])(implicit ev: MonadChronicle[F, E]): F[E Ior A] =
```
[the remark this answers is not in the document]
YuvalItzchakov: That was my initial thought. As I Imagined most of these calls will be via the implicit syntax anyway, but I did it to make sure to align with other MTL instances (perhaps they are called more often via the companion?). Should I remove all of them entirely?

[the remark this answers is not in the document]
LukaJCB: Yeah, I think removing those 5 should be fine :)

tests/src/test/scala/cats/mtl/tests/IorTTests.scala
```diff
@@ -0,0 +1,61 @@
...
+      "IorT[Option, String, ?]",
+      MonadChronicleTests[IorTC[Option, String]#l, String]
+        .monadChronicle[String]
+    )
+
+    checkAll(
+      "Ior[Option, String]",
+      MonadChronicleTests[IorC[String]#l, String]
+        .monadChronicle[String]
+    )
```
LukaJCB: Maybe we could also add a test here for `IorT`s pass-through, so something like `MonadStateTests[IorT[State[String, ?], Int, ?], Int].monadState[Int]`

LukaJCB:
Also a test for lifting through `Ior` and `IorT` to get `MonadChronicle` would be great:

`MonadChronicle[WriterT[Ior[String, ?], Int], String].monadChronicle[String]` 
as well as

`MonadChronicle[EitherT[IorT[Option, String, ?], Int], String].monadChronicle[String]`

LukaJCB:
I also think we could add some laws like:

```scala
dictate(c1) *> dictate(c2) <-> dictate(c1 |+| c2)

dictate(c1) *> confess(c2) <-> confess(c1 |+| c2)
```

core/src/main/scala/cats/mtl/instances/chronicle.scala
```diff
@@ -0,0 +1,68 @@
...
+        ml.layer(under.dictate(c))
+      }
+
+      def confess[A](c: E): M[A] =
+        ml.layer(under.confess(c))
+
+      def materialize[A](fa: M[A]): M[Ior[E, A]] =
+        ml.outerInstance.flatMap(ml.layerControl { nt =>
+          under.materialize(nt(fa))
+        }) {x => x.traverse(ml.restore)(ml.outerInstance) }
```
LukaJCB: This method is really complex and does quite a lot and might therefore be quite inefficient. Maybe it's better to have a different primitive after all and derive `materialize` from there. Sorry for suggesting so many changes.
`def overcome[A](ma: M[A])(f: Either[E, (E, A)] => A): M[A]`. Then this would go away as lifting it would just be ```scala
ml.layerMapK(ma)((new (Inner ~> Inner) {
  def apply[X](fa: Inner[X]): Inner[X] = under.overcome(fa)(f)
})
```

That should also give us the ability to remove the need for `MonadLayerControl` :)

LukaJCB: Disregard this, `under.overcome` would need a function ` Either[E, (E, X)] => X`, so this won't work and we're stuck with this it seems :smile:

LukaJCB:
I'll close and reopen again to trigger travis

YuvalItzchakov:
Looks like Travis is still stuck?

LukaJCB:
Yeah, can you maybe close this PR and open a new one? Not sure what to do in cases like these 😌

YuvalItzchakov:
Closing pull request and opening a new one to try and trigger Travis.

PR #36, opened 2018-07-30
Move layer type classes to `lifting` package

### LINKED ISSUE #33: Move the layer type classes into another package
For all users of this library that don't define their own transformers, they're likely to never see or use these type classes, so having them in the main package feels pretty unnecessary. @SystemFw suggested they go into their own `lifting` package.

denisrosca:
Closes #33

### DIFF
--- a/core/src/main/scala/cats/mtl/lifting/FunctorLayer.scala
+++ b/core/src/main/scala/cats/mtl/lifting/FunctorLayer.scala
@@ -1,5 +1,8 @@
 package cats
 package mtl
+package lifting
+
+import cats.{Functor, ~>}
 
 /**
   * `FunctorLayer[M, Inner]` has the following functionality:
--- a/core/src/main/scala/cats/mtl/lifting/FunctorLayer.scala
+++ b/core/src/main/scala/cats/mtl/lifting/FunctorLayer.scala
@@ -2,8 +2,6 @@ package cats
 package mtl
 package lifting
 
-import cats.{Functor, ~>}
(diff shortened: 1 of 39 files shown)


denisrosca:
Not sure if the tests in `cats.mtl.laws` and `cats.mtl.laws.discipline` should be moved similar to the type classes.

core/src/main/scala/cats/mtl/lifting/FunctorLayer.scala
```diff
@@ -1,5 +1,8 @@
 package cats
 package mtl
+package lifting
+
+import cats.{Functor, ~>}
```
LukaJCB: Aren't these imports redundant since we're already in the `cats` package?

denisrosca:
@LukaJCB Yep, you are correct. Fixed.

LukaJCB:
Looks good to me, thanks very much! :)

LukaJCB:
@andyscott @kailuowang @SystemFw 
We need to think about versioning soon, I think most agreed that this change is definitely worth it, but the question is what's our versioning scheme going forward.

kailuowang:
just realized that somehow I wasn't watching this repo. I vote for just bump the minor for this change since we are still pre 1.0

LukaJCB:
That is my vote as well, should we merge this PR and move the discussion to its own issue (sorry @denisrosca)

SystemFw:
Is this ready to merge then?

PR #40, opened 2018-08-08
MonadChronicle implementation (re-opening for build)

YuvalItzchakov:
See: https://github.com/typelevel/cats-mtl/pull/35


YuvalItzchakov:
@kailuowang I tried adding the additional under GitHub. Perhaps lets try to re-trigger the build?

PR #41, opened 2018-08-08
Add an ApplicativeHandle type class 

### LINKED ISSUE #29: Mtl Class for catching errors
There's `FunctorRaise` for raising errors. Should there also be an mtl class for catching errors?

LukaJCB:
Should fix #29 

/cc @jvliwanag


LukaJCB:
@andyscott if you want we can discuss them in another issue :)

PR #44, opened 2018-08-10
MonadChronicle implementation (try to work around #40 not triggering build)

kailuowang:
* Initial MonadChronicle implementation

* Remove name comment from tests

Formatting

* Improvements after review. Narrowed down the API to just three methods, namely `dictate`, `confess` and `materialize`, all other combinators are now derived from these.

Added `DefaultMonadState` implementation and helper methods to lift into `MonadChronicle` on the companion.

* 1. Several bug fixes in `dictate`, `materialize` and more.
2. Added laws!
3. Added Ior instance for MonadChronicle
4. Added tests for both IorT and Ior instances of MonadChronicle
5. Added more syntax sugar and re-organized implicits to comply with other MTL classes.

* Removed methods from the companion after review.

* Add MonadChronicle for MonadLayerControl

* Add Test for MonadChronicle with Inner Ior and IorT

* Added more tests, fixed imports after merge from master

### DIFF
--- a/core/src/main/scala/cats/mtl/MonadChronicle.scala
+++ b/core/src/main/scala/cats/mtl/MonadChronicle.scala
@@ -1,0 +1,78 @@
+package cats
+package mtl
+
+import cats.data.Ior
+
+trait MonadChronicle[F[_], E] extends Serializable {
+  val monad: Monad[F]
+
+  def dictate(c: E): F[Unit]
+
+  def disclose[A](c: E)(implicit M: Monoid[A]): F[A]
+
+  def confess[A](c: E): F[A]
+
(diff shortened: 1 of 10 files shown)


YuvalItzchakov:
Yay, it runs!

LukaJCB:
🎉

YuvalItzchakov:
Oh man, Travis does not seem to like my commits :X. Looks like it's stuck?

YuvalItzchakov:
Is this mergable as is?

LukaJCB:
This looks great, we should also add some tests for the syntax, but that can also be done in a follow up PR :)

PR #46, opened 2018-08-11
Added monad chronicle syntax tests

YuvalItzchakov:
Missing syntax tests from `MonadChronicle` implementation.


kailuowang:
I think we need to contact GitHub why your PR doesn't trigger webhook. Do your other PRs at other repos work?

YuvalItzchakov:
Yes, haven't had any problems up to now. I'll try and contact them.

YuvalItzchakov:
@kailuowang I contacted GitHub, still no reply (been like a week already) :(. Is there anyway perhaps to look at Travis logs or the likes? Or is this purely a GitHub thing not triggering Travis at all?

kailuowang:
There is nothing on travis, it's like the PR doesn't exist to travis at all. And not much I can see from the github side either.

PR #50, opened 2018-08-20
Monad chronicle tests (last time Travis test)

YuvalItzchakov:


tests/src/test/scala/cats/mtl/tests/Syntax.scala
```diff
@@ -83,6 +83,22 @@ final class Syntax extends BaseSuite {
       val toldC = FunctorTell.tell[WriterTC[Option, String]#l, String]("ha")
       val toldFC = FunctorTell.tellF[WriterTC[Option, String]#l]("ha")
     }
-  }
+    test("MonadChronicle") {
+      val chronicleC: Ior[Int, String] = MonadChronicle.chronicle(Ior.right[Int, String]("w00t"))
```
LukaJCB: This one doesn't seem to work on Scala 2.10.x, you could probably just add a type signature after `chronicle`. :)
Though in the meantime, we should start working on removing 2.10, since Cats doesn't support it anymore :smile:

YuvalItzchakov:
Yeah, I think that's what I'll do. I'd be happy to pick up deprecation of 2.10 if it's agreed upon.

LukaJCB:
It is :) See #45

Issue #47, opened 2018-08-16
Remove FunctorEmpty and TraverseEmpty in conjunction with cats adding them


LukaJCB:
For reference: https://github.com/typelevel/cats/pull/2405

PR #51, opened 2018-08-21
Remove FunctorEmpty and TraverseEmpty

stephen-lazaro:
Resolves issue #47
cf. https://github.com/typelevel/cats/pull/2405/files
One instance cannot be discovered in the RWSTT tests, currently looking into that... Might need to wait on cats [PR2405](https://github.com/typelevel/cats/pull/2405)?


tests/src/test/scala/cats/mtl/tests/ReaderWriterStateTTests.scala
```diff
@@ -26,13 +26,13 @@ class ReaderWriterStateTTests extends StateTTestsBase {
...
     }, FunctionK.id[(L, S, ?)]))
-
+  /**
   checkAll("ReaderWriterStateT[Option, Boolean, Int, String, ?]",
     MonadLayerControlTests[ReaderWriterStateT[Option, Boolean, Int, String, ?], Option, (Int, String, ?)]
       .monadLayerControl[Boolean, Boolean])
   checkAll("MonadLayerControl[ReaderWriterStateT[Option, Boolean, Int, String, ?], Option]",
     SerializableTests.serializable(MonadLayerControl[ReaderWriterStateT[Option, Boolean, Int, String, ?], Option]))
-
+  */
```
stephen-lazaro: Gotta figure out why this instance is no longer found.

LukaJCB:
Thanks for this :) Looks good so far (though that instance not being found is odd)

core/src/main/scala/cats/mtl/instances/all.scala
```diff
@@ -9,6 +9,5 @@ trait AllInstances extends EitherTInstances
   with RaiseInstances with ReaderTInstances
   with LocalInstances with StateInstances
   with StateTInstances with WriterTInstances
-  with EmptyInstances with ReaderWriterStateTInstances
```
LukaJCB: Oh actually it's not odd, you deleted them here :smile:

stephen-lazaro: Cool, fixed now!

stephen-lazaro:
Looks like code coverage went down due to the deletion of some hierarchy stuff that referenced `FunctorEmpty`. Is that acceptable or should I wait on PR2405 and then add it back?

LukaJCB:
I don't think we need to add that back, this should be fine :)

stephen-lazaro:
Cool, sounds good to me! Thanks Luka :)

Issue #26, opened 2018-02-11
Consider to use `InstanceOf` scalaz pattern 


Odomontois:
[This scalaz PR](https://github.com/scalaz/scalaz/pull/1647) targeting well-known ambigous-subtype implicit problem resolving it without repackaging base typeclass instances inside sub-typeclass es.

I wonder if this method is suitable for cats-mtl project. 

PR welcome?

LukaJCB:
This sounds like a good idea to me, I haven't fully understood the pattern yet, but it seems to strike a good balance. We've always said `cats-mtl` is an experimental part of the ecosystem, so I'd love to see a PR :)

SystemFw:
Yes, we should definitely see where this leads.

andyscott:
Agreed! PR is definitely welcome.

LukaJCB:
I wonder if we could get an experimental version of simulacrum with this type of encoding

LukaJCB:
@Odomontois, are you working on this? If not I'd like to give this a shot.

kamilkloch:
@LukaJCB I picked up the task.

LukaJCB:
Closing in favor of #53

PR #52, opened 2018-08-28
Scalaz 8 typeclass encoding. WIP.

kamilkloch:
First draft towards https://github.com/typelevel/cats-mtl/issues/26.
New encodings for:
- [x] ApplicativeAsk
- [x] ApplicativeLocal
- [x] MonadChronicle


core/src/main/scala/cats/mtl/ApplicativeAsk.scala
```diff
@@ -1,5 +1,8 @@
-package cats
-package mtl
+package cats.mtl
+
+import cats.Applicative
+
+import acyclic.skipped
```
LukaJCB: What does this do? Do we need it? :)

kamilkloch: Result of `scalacOptions += "-P:acyclic:force"`

LukaJCB:
Looks good so far, thank you Kamil! 
I wonder if we could get some good benchmarks to really motivate this :)

kamilkloch:
@LukaJCB please take a look into https://gitter.im/scalaz/scalaz, @edmundnoble shed some light.

kamilkloch:
@LukaJCB You are right. This would simplify the code at the cost of adding a highly unlikely potential implicit ambiguity. I was thinking of yet another thing: is there any way to actually inherit (not aggregate) from cats' typeclasses? I mean `ApplicativeAsk[F[_], E] extends Applicative[F]` ;)

SystemFw:
>  is there any way to actually inherit (not aggregate) from cats' typeclasses? I mean ApplicativeAsk[F[_], E] extends Applicative[F]

No, that would bring you the ambiguous implicit problem `scato` solves, since e.g. `F[_]: ApplicativeAsk : ApplicativeLocal` would have two `Applicative`s

LukaJCB:
We could use the `instanceOf` pattern for `ApplicativeAsk` and `Applicative`, though I'm sceptical that would buy us all too much.

LukaJCB:
> We can do that, but it kinda constrains to not add any more leaves right? Unsure if it's a problem or not, practically

@SystemFw if we do add more leaves, we can just `scato` for that leaf again.

LukaJCB:
> F[_]: ApplicativeAsk : ApplicativeLocal would also fail if ApplicativeLocal would plainly extend ApplicativeAsk, as suggested by @LukaJCB . I was more thinking of using the hidden inheritance pattern from scalaz also for Applicative, for example
ApplicativeAskClass[F[_], E] extends ApplicativeClass[F]
(i.e., switch fully to the ineritance mode by adding some instanceOf plumbing, just an idea)

I'm not sure if this buys us a lot, unless we provide an implicit conversion from e.g. `ApplicativeAsk` to `Applicative`. Because right now, `cats-mtl` has you write code like this:

```scala
def foo[F[_]: Monad: ApplicativeAsk[?[_], Config]: F[A] = ???
```

LukaJCB:
Yes, I think it looks good this way :+1:

kamilkloch:
If we are to proceed, shouldn't we do it in a new PR?

LukaJCB:
Sure Sounds good to me :)

LukaJCB:
I'll close this for now, we can always reopen if we reconsider :)

Issue #53, opened 2018-08-29
Consider inheritance-based typeclass encoding


kamilkloch:
As a simpler variant of https://github.com/typelevel/cats-mtl/issues/26.
See discussion in https://github.com/typelevel/cats-mtl/pull/52.

PR #54, opened 2018-08-29
Plain typeclass inheritance encoding.

kamilkloch:
Work towards https://github.com/typelevel/cats-mtl/issues/53.

- [x] `ApplicativeLocal` extends `ApplicativeAsk`
- [x] `FunctorListen` extends `FunctorTell`
- [x] `ApplicativeHandle` extends `ApplicativeRaise`


core/src/main/scala/cats/mtl/instances/empty.scala
```diff
@@ -328,9 +320,37 @@ trait EmptyInstances1 {
 
   implicit def traverseEmptyLiftEitherT[M[_], E](implicit under: TraverseEmpty[M]): TraverseEmpty[EitherTC[M, E]#l] = {
     new TraverseEmpty[EitherTC[M, E]#l] {
-      override val functorEmpty: FunctorEmpty[EitherTC[M, E]#l] = functorEmptyLiftEitherT(under.functorEmpty)
+
+      //implicit def functorEmptyLiftEitherT[M[_], E](implicit under: FunctorEmpty[M]): FunctorEmpty[EitherTC[M, E]#l] = {
```
LukaJCB: `scalastyle` wants you to put a space here before the comment body :)

kamilkloch: My bad, cleaned up.

PR #57, opened 2018-09-08
Remove JDK 7 and Scala 2.10 support

YuvalItzchakov:
Fixes https://github.com/typelevel/cats-mtl/issues/45


kailuowang:
Thanks. I believe those settings can just be removed.

project/Docs.scala
```diff
@@ -47,8 +37,7 @@ object Docs {
       "gray-lighter" -> "#F4F3F4",
       "white-color" -> "#FFFFFF"),
     autoAPIMappings := true,
-    unidocProjectFilter in(ScalaUnidoc, unidoc) :=
-      inProjects(docsSourcesAndProjects(scalaVersion.value)._2: _*),
+    unidocProjectFilter in(ScalaUnidoc, unidoc) := inProjects(Seq.empty[ProjectReference]: _*),
```
YuvalItzchakov: @kailuowang @LukaJCB I'm not sure this is necessary anymore, just wasn't sure. Thoughts?

project/Docs.scala
```diff
@@ -10,20 +10,10 @@ import com.github.tkawachi.doctest.DoctestPlugin.autoImport._
...
-  }
-
   lazy val docsMappingsAPIDir = settingKey[String]("Name of subdirectory in site target directory for api docs")
 
   lazy val javadocSettings = Seq(
-    sources in(Compile, doc) := {
-      if (docsSourcesAndProjects(scalaVersion.value)._1) (sources in(Compile, doc)).value
-      else Nil
-    }
+    sources in(Compile, doc) := (sources in(Compile, doc)).value
```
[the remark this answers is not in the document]
kailuowang: You are right.

project/Docs.scala
kailuowang: This setting should be removed. Otherwise if it will disable the scaladoc for the project

project/Docs.scala
```diff
@@ -47,8 +33,6 @@ object Docs {
       "gray-lighter" -> "#F4F3F4",
       "white-color" -> "#FFFFFF"),
     autoAPIMappings := true,
-    unidocProjectFilter in(ScalaUnidoc, unidoc) :=
```
YuvalItzchakov: @kailuowang Removing this line seems to be generating compilation errors during the build:

```
info] Main Scala API documentation to /home/travis/build/typelevel/cats-mtl/docs/target/scala-2.11/unidoc...
[error] /home/travis/build/typelevel/cats-mtl/core/src/main/scala/cats/mtl/ApplicativeAsk.scala:24: ApplicativeAsk is already defined as trait ApplicativeAsk
[error] trait ApplicativeAsk[F[_], E] extends Serializable {
[error]       ^
[error] /home/travis/build/typelevel/cats-mtl/core/src/main/scala/cats/mtl/ApplicativeAsk.scala:32: ApplicativeAsk is already defined as object ApplicativeAsk
```
Full log: https://travis-ci.org/typelevel/cats-mtl/jobs/426334463

kailuowang: try set it to `Seq(coreJVM, lawsJVM)`?

YuvalItzchakov: The previous comment seemed to be saying ghaty an empty sequence is equivalent to setting all the projects: kernelJVM, coreJVM, freeJVM

kailuowang: Those are copied from Cats. 
I'd say let's copy from the settings in cats now. 
https://github.com/typelevel/cats/blob/master/build.sbt#L198
in our case it's just coreJVM and lawsJVM

YuvalItzchakov: Ok, great. Will do.

YuvalItzchakov: @kailuowang So all example settings I've seen (i.e. in cats), the definition for the docs resides inside the `build.sbt`. Specifically here, we're inside another `Docs` object, which has no access to `coreJVM` or `lawsJVM` values defined.

Should I move this setting into `build.sbt` directly? Or perhaps I'm missing something here.

kailuowang: oh I didn't realize that they are in a different file.  I would propose that we move this back to build.sbt.  I am more in the everything in `build.sbt` camp.

LukaJCB: +1 to "everything in `build.sbt`"

YuvalItzchakov:
I've moved `Docs` into `build.sbt`, perhaps we should open a separate issue to consolidate the rest?

build.sbt
```diff
@@ -27,22 +29,68 @@ val core = crossProject.crossType(CrossType.Pure)
...
+
+lazy val docSettings = Seq(
+  micrositeName := "Cats MTL",
+  micrositeDescription := "Companion library to cats providing monad transformers",
+  micrositeAuthor := "Typelevel contributors",
+  micrositeHighlightTheme := "atom-one-light",
+  //  micrositeHomepage := "http://typelevel.org/cats",
+  micrositeBaseUrl := "cats-mtl",
+  micrositeDocumentationUrl := "api",
+  micrositeGithubOwner := "edmundnoble",
```
andyscott: Might as well adjust this to the same as cats: `micrositeGithubOwner := "typelevel"`

PR #63, opened 2018-09-27
WIP: Microsite

LukaJCB:


LukaJCB:
I think this is ready for review now.
It's not fully complete, but I think it'd be great if we could merge this soon and have a baseline website to work with :)

LukaJCB:
@larsrh, @kailuowang it would be great if one of you could help me setup the website at `typelevel.org/cats-mtl` once this is merged :)

larsrh:
@LukaJCB No need for setup, this will be automatic. You only have to run `publishMicrosite`.

COPYING
```diff
@@ -1,4 +1,4 @@
-cats-mtl Copyright (c) 2017 Edmund Noble.
+cats-mtl Copyright (c) 2017 Cats-mtl Contributors.
```
andyscott: Do we have permission to change the copyright ownership?

LukaJCB: Great question, I have no idea, I just wanted to make it consistent with `cats` and `cats-effect`.

larsrh: Is that a problem? Surely it is a superset now.

LukaJCB:
Shall we go ahead and merge this then? :)

larsrh:
I'll also update #64 after that. Surely we should be able to bump to sbt 1.x.

larsrh:
> Yes, we should release a new version after that as well

In that case I suggest:
* `0.3.1` release immediately
* run `publishMicrosite` after that

LukaJCB:
Agreed, though let's call it `0.4.0` :)

PR #64, opened 2018-09-29
sbt 1.0

larsrh:
Updates all plugins and sbt to 1.x.

There are still some weird errors in here.

```
[error] (node:2872) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[error] (node:2891) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[error] /home/lars/proj/typelevel/cats-mtl/tests/src/test/scala/cats/mtl/tests/package.scala:1:1: 
[error] Found names but no class, trait or object is defined in the compilation unit.
[error] The incremental compiler cannot record the dependency information in such case.
[error] Some errors like unused import referring to a non-existent class might not be reported.
[error]     
[error] package cats
[error] ^
[error] one error found
```


larsrh:
Locally, this appears to run fine. I was a bit concerned about messages from ScalaTest that no tests were run but those aren't new. Total number of tests executed match before and after this PR.

PR #65, opened 2018-09-30
Delete unused test package object

LukaJCB:
Should unblock #64


larsrh:
Oh wow. Why didn't this fail before?

LukaJCB:
No idea :smile:

PR #70, opened 2018-09-30
Mention the 'primary monad transformer' for each of the type classes

LukaJCB:
Does this make sense as a clarification? Maybe there's another term. It feels *kinda* wrong, but the information is still really good to know, I think 🤔


SystemFw:
I'm going to merge for now, since the info is useful. If we can think of a better term we can update later

Issue #71, opened 2018-09-30
Add revised MonadLog?


vlovgr:
If we adapt the `MonadLog` type class presented in [IO & Logging Capabilities](https://blog.vlovgr.se/post/2018-09-24-io-and-logging-capabilities/#introducing-monadlog) to fit as an extension to `FunctorListen`, we end up with the following. We've replaced `log` with `tell` and `extract` is replaced by `listen`.

```scala
trait MonadLog[F[_], L] extends FunctorListen[F, L] {
  val monad: Monad[F]

  val monoid: Monoid[L]

  def clear[A](fa: F[A]): F[A]

  def flush[A](fa: F[A])(f: L => F[Unit]): F[A]
}
```

Personally, I don't think it's the right way to go:

- we're extending from a `Functor` hierarchy, while we require `Monad`,
- this makes us require not one, but two additional constraints, with `Monoid`,
- logs are forced to take a form like `F[(A, L)]`, and cannot be stored separately.

Instead, I propose we adapt `MonadLog` as an extension to `MonadState`. In fact, any `MonadState[F, L]` can be a `MonadLog[F, L]` with a `Monoid[L]` instance, meaning we can reuse the existing `MonadState` instances. Such an adapted version of `MonadLog` could be defined as follows.

```scala
trait MonadLog[F[_], L] extends MonadState[F, L] with Serializable {
  val monoid: Monoid[L]

  def log(l: L): F[Unit]

  def clear: F[Unit]

  def flush(f: L => F[Unit]): F[Unit]
}
```

Default implementation could be defined as follows.

```scala
trait DefaultMonadLog[F[_], L] extends DefaultMonadState[F, L] with MonadLog[F, L] {
  def log(l: L): F[Unit] =
    modify(monoid.combine(_, l))

  def clear: F[Unit] =
    set(monoid.empty)

  def flush(f: L => F[Unit]): F[Unit] =
    monad.flatMap(monad.flatMap(get)(f))(_ => clear)
}
```

Laws could be defined as follows.

```scala
trait MonadLogLaws[F[_], L] extends MonadStateLaws[F, L] {
  implicit val logInstance: MonadLog[F, L]
  implicit val monoid: Monoid[L] = logInstance.monoid

  import logInstance._
  import monad.{pure, unit}
  import monoid.empty

  def logThenClearDoesNothing(l: L) =
    log(l) *> clear <-> unit

  def logThenGetReturnsLog(l: L) =
[...]

PR #82, opened 2018-10-04
Add ApplicativeCensor

LukaJCB:
This plays into #71, it's not quite the `ApplicativePass` I proposed there, but a slightly weaker `ApplicativeCensor` that cannot use the `F` context to build the function. This was done, so that it can be lifted through transformer stacks, which wasn't possible otherwise.


SystemFw:
Just a heads up that I won't have time to review this today, in all likelihood

LukaJCB:
No worries, take your time, no rush at all :)

Issue #80, opened 2018-09-30
Explore a submodule for more performant instances.


LukaJCB:
As many have said before, stacking multiple monad transformers can be quite inefficient, it'd be cool if we had some kind of story for supporting performant instances for different combination of mtl classes.
Oleg has already done some work in that regard with `meow-mtl` and at the start of the year, I've also played around with specialized instances at [cats-mtl-special](https://github.com/LukaJCB/cats-mtl-special).

iravid:
Happy to contribute https://github.com/iravid/irwst :)

LukaJCB:
@iravid That looks really cool, I think it'd be a great fit :)

PR #86, opened 2018-10-12
WIP: New special and bench modules + StateIO

LukaJCB:
Should address #80 for now.

StateIO uses a `Ref` to make state updates more performant.
A simple benchmark included in this PR shows this so far:

```
[info] Benchmark                              Mode  Cnt    Score    Error  Units
[info] StateBench.leftAssociatedBindSpecial  thrpt   10  297.506 ± 13.396  ops/s
[info] StateBench.leftAssociatedBindStateT   thrpt   10  107.687 ±  3.500  ops/s
```

next step might be to newtype it, to avoid boxing.


LukaJCB:
Hmm, turning it into a newtype somehow degregaded performance of `flatMap` a bit

```
[info] Benchmark                              Mode  Cnt    Score    Error  Units
[info] StateBench.leftAssociatedBindSpecial  thrpt   10  234.863 ± 12.094  ops/s
[info] StateBench.leftAssociatedBindStateT   thrpt   10  106.308 ±  3.421  ops/s
```

PR #95, opened 2018-12-08
Fix MonadChronicle signatures and ops syntax

catostrophe:
I've found some issues in the `MonadChronicle` typeclass:

- [x] Signatures of a few functions were copied from Haskell as is, while in `cats` another order of arguments is preferred. (fixed)
- [x] Some ops functions wrongly extend `F[A]` value, which is not used. There should be several ops extensions for different types. (fixed)
- [x] (**declined**) `disclose` requires `Monoid[A]` instance, which is too strong. Actually it is enough to have `Empty[A]` (as in Haskell implementation). But I haven't changed it and would like to discuss it at first

### DIFF
--- a/core/src/main/scala/cats/mtl/syntax/chronicle.scala
+++ b/core/src/main/scala/cats/mtl/syntax/chronicle.scala
@@ -6,21 +6,11 @@ import cats.data.Ior
 
 trait ChronicleSyntax {
   implicit def toChronicleOps[F[_], A](fa: F[A]): ChronicleOps[F, A] = new ChronicleOps[F, A](fa)
+  implicit def toChronicleIdOps[E](e: E): ChronicleIdOps[E] = new ChronicleIdOps[E](e)
+  implicit def toChronicleIorOps[A, E](ior: E Ior A): ChronicleIorOps[A, E] = new ChronicleIorOps[A, E](ior)
 }
 
 final class ChronicleOps[F[_], A](val fa: F[A]) extends AnyVal {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/mtl/syntax/chronicle.scala
```diff
@@ -30,18 +20,34 @@ final class ChronicleOps[F[_], A](val fa: F[A]) extends AnyVal {
...
+    chronicle.retcon(fa)(cc)
   }
+}
+
+final class ChronicleIdOps[E](val e: E) extends AnyVal {
+  def dictate[F[_]](implicit chronicle: MonadChronicle[F, E]): F[Unit] = {
+      chronicle.dictate(e)
+    }
+
+  def disclose[F[_], A](implicit chronicle: MonadChronicle[F, E], monoid: Monoid[A]): F[A] = {
```
catostrophe: Originally `disclose` needs `Default` (or cats' `Empty`) intance:
https://hackage.haskell.org/package/these-0.7.5/docs/Control-Monad-Chronicle.html#v:disclose

catostrophe: Changing `Monoid` bound to `Empty` only adds `alleycats` dependency, but doesn't break the existing code, since `Empty` is derived from `Monoid`.

LukaJCB: I think adding an `alleycats` dependency is a non-starter personally. I think it's fine to break with Haskell in this case. :)

LukaJCB:
Cool :)

PR #97, opened 2018-12-16
Add ApplicativeAsk.const

kubukoz:
This PR adds a handy way to create an ApplicativeAsk instance that always answers with the same value.
Useful for when you optimize your program by dropping Reader from your transformer stack and rely on a custom Applicative instance that just wraps over a constant value known at the point of creating it.


kubukoz:
@SystemFw can I ask for another thumbs up? (and maybe a merge) :D

kubukoz:
Thanks!

PR #105, opened 2019-02-03
simplify build

larsrh:
Currently, the 2.11 build doesn't work locally for me. I'm simplifying some things to try to figure out what's wrong.


build.sbt
```diff
@@ -7,38 +7,96 @@ organization in ThisBuild := "org.typelevel"
...
 
-addCommandAlias("validateJS", ";catsMtlJS/compile;testsJS/test;js/test")
+addCommandAlias("validateJS", ";testsJS/compile;testsJS/test")
 
 addCommandAlias("validate", ";clean;validateJS;validateJVM")
 
 addCommandAlias("gitSnapshots",
                 ";set version in ThisBuild := git.gitDescribedVersion.value.get + \"-SNAPSHOT\"")
 
-sbtPlugin := true
```
larsrh: This line was what caused the issue. *\*shakes fist\**

larsrh:
I have a follow-up PR for 2.13.0-M5, and then we can get a release going.

build.sbt
kubukoz: Damn, this line...

larsrh:
```scala
val total_hours_wasted_here = 3
```

PR #106, opened 2019-02-03
Scala 2.13.0-M5

larsrh:


larsrh:
Blockey by tpolecat/tut#246 or changing the build

larsrh:
I'll send another (WIP) PR after this one so we don't forget to revert the workaround (1e89758).

PR #108, opened 2019-02-13
[WIP] Revert "workaround for tut and Scala 2.13.0-M5"

larsrh:
See #106. Will fail to run until tut releases a fix.


larsrh:
I'm closing this because it seems unlikely to be fixed. Instead, we should probably migrate to mdoc.

PR #117, opened 2019-05-25
Update kind-projector to 0.10.0

ybasket:
Now has new `org.typelevel` organization, that's why scala-steward won't pick it up.


larsrh:
This change has already been merged in another PR.

PR #119, opened 2019-06-17
POC: Implement functor/contravariant/invariant for some TCs

kubukoz:
I had a need for `Functor` on `ApplicativeAsk`, so I just went with it and looked at other possible implementations.

TODO:
- [x] instances for other type classes
- [ ] law tests, obviously

Let me know if this looks like a nice thing to have :)

### DIFF
--- a/core/src/main/scala/cats/mtl/Ask.scala
+++ b/core/src/main/scala/cats/mtl/Ask.scala
@@ -78,6 +78,15 @@ private[mtl] trait AskInstances extends LowPriorityAskInstances {
       implicit F: Monad[F],
       L: Monoid[L]): Ask[RWST[F, E, L, S, *], E] =
     Local.baseLocalForRWST[F, E, L, S]
+
+  implicit def catsFunctorForAsk[F[_]]: Functor[Ask[F, *]] =
+    new Functor[Ask[F, *]] {
+      def map[A, B](fa: Ask[F, A])(f: A => B): Ask[F, B] =
+        new Ask[F, B] {
+          val applicative: Applicative[F] = fa.applicative
+          def ask[E2 >: B]: F[E2] = applicative.map(fa.ask)(f)
+        }
+    }
 }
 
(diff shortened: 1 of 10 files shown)


djspiewak:
I can't quite decide if this is nice to have or not. :-D It allows you to derive new instances really easily, but it also feels so bizarre for a typeclass to form an instance of another typeclass.

kubukoz:
@djspiewak I finally revisited this, had to basically start over because of the new naming/hierarchy.

I added a couple examples of the approach I'd take to law-test these instances - the tough part is the equality definitions of the type classes (`Eq[Raise]` etc.).

I used [the instances of `Eq[Semilattice]` and friends](https://github.com/typelevel/cats/blob/v2.3.1/laws/src/main/scala/cats/laws/discipline/Eq.scala#L91-L96) as inspiration (`Eq.by` and comparing all the functions in the typeclass), but that only worked as long as the methods had no type parameters - the moment a type param appears, it becomes more complicated.

For `Listen`, which has the `listen` method, which has a type parameter, I tried to do some polymorphic magic with `skolems`, but that seemed to hang the compiler or something 😅 so in the end, I went with a less universal approach, adding a new type parameter to the Eq instance function:

```scala
implicit def catsEqForListen[F[_], A, E](
    implicit eqTell: Eq[A => F[Unit]],
    eqListen: Eq[F[E] => F[(E, A)]]): Eq[Listen[F, A]] = {
  Eq.by { instance =>
    (
      instance.tell _,
      instance.listen[E] _
    )
  }
```

Then it needs another method to forward to this one, because `E` could never be inferred otherwise - but it works. What do you think?

Another question I have is whether these `Eq` instances should be in the core module (with the type classes they're defined for) or in the `laws` module, like the Cats instances.

Finally, instances of other TCs embedded as fields (e.g. `Functor` in `Tell`). For now, I decided to ignore these but maybe I shouldn't.

PR #132, opened 2019-07-12
Initial 2.13 port

kaishh:
Fix https://github.com/typelevel/cats-mtl/issues/128

Not so pretty, since I removed catalysts `isJvm` check and copied `Arbitrary ~> ExhaustiveCheck` instance from cats-tagless, and that instance causes Stream deprecation works, but still at least now it passes `validate` and I think that's most of the porting work timewise.


kaishh:
@LukaJCB Thanks! Do you think you could release a new version in some time? Thank you

LukaJCB:
Yes, I agree, we should do this soon :)

kaishh:
Thank you!

PR #145, opened 2019-09-10
Updating things, a little build cleanup, ready for release

djspiewak:
I think this is all that is necessary to get a 2.0-compatible release out the door.


djspiewak:
I don't really understand why the build failed. I blame gremlins.

PR #159, opened 2019-10-04
dependency bumps

larsrh:


larsrh:
Blocked on scoverage/sbt-scoverage#295

larsrh:
This still doesn't work, but I have no idea why.

fthomas:
I think you need to bump sbt-mdoc to 1.3.6: https://github.com/scalameta/mdoc/releases/tag/v1.3.6

larsrh:
Thanks, that did the trick!

PR #185, opened 2020-01-23
Update discipline-scalatest

travisbrown:
Supersedes #184, which fails because there is a breaking change in this RC.


travisbrown:
The build failure is a 403 and looks like rate limiting or just general Maven Central weirdness, and I don't have the permissions needed to restart that job.

PR #208, opened 2020-04-05
Add EitherT, OptionT, ReaderT, WriterT covariant transformers

LukaJCB:
Early draft for covariant transformers. Still missing pretty much all the instances, but otherwise seem to work fine.


djspiewak:
I think this is roughly superseded by #234

PR #209, opened 2020-04-08
Rewrite all the things!

djspiewak:
This PR is a very work in progress dump of the current state of one of my branches. It substantially rewrites *everything*. The core structure of the typeclasses remain present, but I removed `MonadControlLayer` (and friends) because they're unsound outside of a very limited set of datatypes (e.g. they're sound for `EitherT` and `StateT`, but not for `ContT`, `FreeT`, any hypothetical `StreamT`, etc). Even worse, this unsoundness is not captured by the types at all, so it's very easy to write invalid `MonadControlLayer` instances which only fail in certain structural contexts (e.g. the naive `FreeT` instance is broken, but most ScalaCheck generators are not robust enough to catch it even if the laws captured the problem, which they don't). In its place is a more restricted typeclass: `MonadPartialOrder`.

The idea for this is originally from @nuttycom, and for whatever reason it never saw wide adoption despite being introduced into Scalaz pre-7.0. It's effectively a generalization of `Hoist`, where you can talk about how effects can (and cannot) nest within each other. Unlike `MonadControlLayer`, it doesn't attempt to handle the *deconstruction* side of the equation, only the *reconstruction* side. This still results in significantly reduced boilerplate in implementations (you can see good examples of this within this PR) without introducing the soundness limitations.

More importantly, in this rewrite, `MonadPartialOrder` is a *convenience*, not a necessity. `MonadControlLayer` encodings require all participating datatypes to define an instance, otherwise they do not work inductively over *any* other datatypes. `MonadPartialOrder` is optional and simply makes it easier to define the boilerplate.

Oh, added side-benefit: there is no longer a need to import `cats.mtl.implicits._` since everything can be tucked into companion objects. So that's nice.

Speaking of boilerplate, there is objectively more of it now than before, but that boilerplate is (as ever) confined to datatype and typeclass authors. More importantly, there is *control* over that boilerplate, making it possible for datatype authors to control the semantics of complex compositions, such as `ContT` and `StateT`. In other words, I think the boilerplate carries its on weight.


nuttycom:
Credit for MonadPartialOrder goes to Oleg, not me! I just learned about it from his work on iteratees.

LukaJCB:
Oh boy, a bunch of scalastyle failures, I wonder if it's worth fixing them instead of just going with scalafmt.

djspiewak:
@LukaJCB IMO let's just rip out scalastyle. Can add scalafmt either now or later.

LukaJCB:
Closing in favor of #215

PR #215, opened 2020-05-14
Rewrite all the things: Part 2

LukaJCB:


core/src/main/scala/cats/mtl/ApplicativeHandle.scala
```diff
@@ -1,34 +1,177 @@
...
+
+      def raise[A](e: E): IorT[F, E, A] = IorT.leftT(e)
+
+      def handleWith[A](fa: IorT[F, E, A])(f: E => IorT[F, E, A]): IorT[F, E, A] = IorT(F.flatMap(fa.value) {
+        case Ior.Left(e) => f(e).value
+        case e @ _ => F.pure(e)
+      })
+    }
+  
+  implicit final def handleIor[E](implicit E: Semigroup[E]): ApplicativeHandle[Ior[E, *], E] =
```
ritschwumm, line 113: the naming for these instances seems to be quite inconsistent:
there is handleIor,  applicativeHandleKleisli, and applicativeLocalForKleisli.

djspiewak: Would definitely like to be more consistent here. Should probably follow the same convention as cats-core.

core/src/main/scala/cats/mtl/FunctorListen.scala
```diff
@@ -1,42 +1,172 @@
...
+    def functor = Functor[RWST[F, E, L, S, *]]
+    def tell(l: L) = RWST.tell(l)
+    def listen[A](fa: RWST[F, E, L, S, A]) =
+      RWST applyF {
+        fa.runF map { f => (e: E, s: S) =>
+          f(e, s) map {
+            case (l, s, a) => (l, s, (a, l))
+          }
+        }
+      }
```
djspiewak, line 102: Starting to think this construct is just hitting a bug in scalafmt…

PR #224, opened 2020-05-31
Removed and replaced most of the old crufty build things

djspiewak:
Redoing pretty much all of the build infrastructure. Now based on sbt-spiewak, which gives us nice snapshots and compiler options and stuff (though I think the 2.13 stuff may need to be redone). Also moved to sbt-github-actions. Updated scala-js. You know, all that stuff.

Oh, I also renamed "cats-mtl-core" to just "cats-mtl" in the artifact names, because I really dislike the "-core" convention and I think it's mostly useless most of the time it is employed.

Fixes #147

PR #228, opened 2020-06-01
Added @implicitNotFound everywhere

djspiewak:
Depends on #224 for sanity. An example error:

```
scala> import cats.mtl._
import cats.mtl._

scala> def foo[F[_], A](fa: F[A])(implicit fask: ApplicativeAsk[F, Int]) = fask.ask
foo: [F[_], A](fa: F[A])(implicit fask: cats.mtl.ApplicativeAsk[F,Int])F[Int]

scala> foo(Some(42))
<console>:16: error: Could not find an implicit instance of ApplicativeAsk[Some, Int]. If you have a
value of type Int in scope, or a way of computing one, you may want to construct
a value of type Kleisli for this call-site, rather than type Some. An example type:

  Kleisli[Some, Int, *]

If you do not have an Int or a way of getting one, you should add
an implicit parameter of this type to your function. For example:

  (implicit fask: ApplicativeAsk[Some, Int}])

       foo(Some(42))
          ^

```


djspiewak:
This has effectively superseded by #230 since I missed `Censor` entirely.

PR #230, opened 2020-06-01
Renamed all the things

### LINKED ISSUE #221: Revisit naming conventions
The names here (e.g. `FunctorTell`) are all *super* arbitrary. There's very little prior art here, since most other MTL systems use encodings that avoid orphans (e.g. Haskell and Scalaz 8), so the cats-mtl calculus is pretty unique. What we should ask ourselves is whether or not the names, as they are currently chosen, appropriately guide users to the meaning and use-cases.

djspiewak:
Renamed all the classes to remove their base dependent upstream class from the name. For the most part, this is just an exercise in concision, since I think "`FunctorAsk`" doesn't really add anything useful over "`Ask`". Subjectively, I think the new names "read" a little better when thinking about things in terms of capabilities.

I left `Chronicle` as, uh, `Chronicle` specifically because it's a carbon-copy of `Control.Monad.Chronicle` in Haskell, and I'd rather not go off inventing names for stuff that has already been named.

Depends on #228 
Fixes #221


djspiewak:
Just realized I forgot to rename some of the files. I'll take care of that quickly

PR #234, opened 2020-06-09
Bake variance into Ask, Raise and Tell

### LINKED ISSUE #219: Ensure all classes implicitly narrow/widen as appropriate
For example, the `Ask` classes should implicitly widen, while the `Error` classes should implicitly narrow. The former is equivalent to making `Kleisli` contravariant, while the latter is equivalent to making `EitherT` covariant, but critically implicit encodings of this work even on polymorphic types. Quick example:

```scala
object ApplicativeAsk {
  implicit def widen[F[_], A, B >: A](implicit F: ApplicativeAsk[F, A]): ApplicativeAsk[F, B] = ...
```

LukaJCB:
Should fix #219

### DIFF
--- a/core/src/main/scala/cats/mtl/Ask.scala
+++ b/core/src/main/scala/cats/mtl/Ask.scala
@@ -43,10 +43,10 @@ import scala.annotation.implicitNotFound
   */
 @implicitNotFound(
   "Could not find an implicit instance of Ask[${F}, ${E}]. If you have a\nvalue of type ${E} in scope, or a way of computing one, you may want to construct\na value of type Kleisli for this call-site, rather than type ${F}. An example type:\n\n  Kleisli[${F}, ${E}, *]\n\nIf you do not have an ${E} or a way of getting one, you should add\nan implicit parameter of this type to your function. For example:\n\n  (implicit fask: Ask[${F}, ${E}}])\n")
-trait Ask[F[_], E] extends Serializable {
+trait Ask[F[_], +E] extends Serializable {
   def applicative: Applicative[F]
 
-  def ask: F[E]
+  def ask[E2 >: E]: F[E2]
 
   def reader[A](f: E => A): F[A] = applicative.map(ask)(f)
 }
@@ -56,7 +56,7 @@ private[mtl] trait AskForMonadPartialOrder[F[_], G[_], E] extends Ask[G, E] {
   val F: Ask[F, E]
 
   override def applicative = lift.monadG
-  override def ask = lift(F.ask)
+  override def ask[E2 >: E] = lift(F.ask)
 }
 
 private[mtl] trait LowPriorityAskInstances extends LowPriorityAskInstancesCompat {
@@ -89,7 +89,7 @@ object Ask extends AskInstances {
   def const[F[_]: Applicative, E](e: E): Ask[F, E] =
     new Ask[F, E] {
       val applicative: Applicative[F] = Applicative[F]
-      val ask: F[E] = applicative.pure(e)
+      def ask[E2 >: E]: F[E2] = applicative.pure(e)
     }
 
   def ask[F[_], E](implicit ask: Ask[F, E]): F[E] =
--- a/core/src/main/scala/cats/mtl/syntax/raise.scala
+++ b/core/src/main/scala/cats/mtl/syntax/raise.scala
@@ -23,7 +23,7 @@ trait RaiseSyntax {
 }
 
 final class RaiseOps[E](val e: E) extends AnyVal {
-  def raise[F[_], A](implicit raise: Raise[F, _ >: E]): F[A] =
+  def raise[F[_], A](implicit raise: Raise[F, E]): F[A] =
     raise.raise(e)
 }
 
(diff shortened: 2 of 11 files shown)


core/src/main/scala/cats/mtl/syntax/raise.scala
```diff
@@ -23,7 +23,7 @@ trait RaiseSyntax {
 }
 
 final class RaiseOps[E](val e: E) extends AnyVal {
-  def raise[F[_], A](implicit raise: Raise[F, _ >: E]): F[A] =
+  def raise[F[_], A](implicit raise: Raise[F, E]): F[A] =
```
djspiewak, line 26: 🎉

tests/src/test/scala/cats/mtl/tests/Syntax.scala
```diff
@@ -59,6 +68,12 @@ final class Syntax extends BaseSuite {
       def fb[F[_]: Raise[?[_], EE], E <: EE, EE](e: E): F[E] = e.raise[F, E]
       val faC: Either[String, Int] = Raise.raise[EitherC[String]#l, String, Int]("ha")
       val faeC: Either[String, Nothing] = Raise.raiseF[EitherC[String]#l]("ha")
+
+      def bar[F[_]](implicit F: Raise[F, Bar]) =
+        F.raise(Bar(404))
+
+      def foo[F[_]](implicit F: Raise[F, Foo]): F[Nothing] =
+        bar[F]
```
djspiewak, line 76: Another case that would be nice to add is to force the compiler to do subtyping along both axes:

```scala
def foo[F[_]](implicit F: Handle[F, Throwable]) = 
  new RuntimeException("yo").raise[F]
```

tests/src/test/scala/cats/mtl/tests/Syntax.scala
```diff
@@ -82,6 +97,12 @@ final class Syntax extends BaseSuite {
       val tupled: WriterT[Option, String, Unit] = ("ha", ()).tuple[WriterTC[Option, String]#l]
       val toldC = Tell.tell[WriterTC[Option, String]#l, String]("ha")
       val toldFC = Tell.tellF[WriterTC[Option, String]#l]("ha")
+
+      def bar[F[_]](implicit F: Tell[F, Bar]) =
+        F.tell(Bar(404))
+
+      def foo[F[_]](implicit F: Tell[F, Foo]): F[Unit] =
+        bar[F]
```
djspiewak, line 105: I think in general we really want to check the syntax resolution more than the raw subtyping. I'm pretty sure most of it will work, but I'm worried about that more than I am relatively simple subsumption cases like this one.

core/src/main/scala/cats/mtl/Ask.scala
```diff
@@ -43,10 +43,10 @@ import scala.annotation.implicitNotFound
   */
 @implicitNotFound(
   "Could not find an implicit instance of Ask[${F}, ${E}]. If you have a\nvalue of type ${E} in scope, or a way of computing one, you may want to construct\na value of type Kleisli for this call-site, rather than type ${F}. An example type:\n\n  Kleisli[${F}, ${E}, *]\n\nIf you do not have an ${E} or a way of getting one, you should add\nan implicit parameter of this type to your function. For example:\n\n  (implicit fask: Ask[${F}, ${E}}])\n")
-trait Ask[F[_], E] extends Serializable {
+trait Ask[F[_], +E] extends Serializable {
   def applicative: Applicative[F]
 
-  def ask: F[E]
+  def ask[E2 >: E]: F[E2]
```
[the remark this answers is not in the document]
LukaJCB: I think the that'd probably be widened to `F[Any]` 😬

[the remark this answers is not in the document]
djspiewak: Surprisingly not!

```
sbt:cats-mtl> console
[info] Compiling 7 Scala sources to /Users/daniel/Development/Scala/cats-mtl/core/.jvm/target/scala-2.13/classes ...
[info] Compiling 8 Scala sources to /Users/daniel/Development/Scala/cats-mtl/core/.jvm/target/scala-2.13/classes ...
[info] Starting scala interpreter...
Welcome to Scala 2.13.2 (OpenJDK 64-Bit Server VM GraalVM CE 20.1.0, Java 1.8.0_252).
Type in expressions for evaluation. Or try :help.

scala> import cats.mtl._
import cats.mtl._

scala> def foo[F[_]](implicit F: Ask[F, String]) = F.ask
def foo[F[_]](implicit F: cats.mtl.Ask[F,String]): F[String]

scala> 
```

tests/src/test/scala/cats/mtl/tests/Syntax.scala
djspiewak: Actually I discovered this case already exists in the syntax tests, so we're good.

core/src/main/scala/cats/mtl/Ask.scala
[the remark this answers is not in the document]
LukaJCB: Oh thank the scalac gods.

djspiewak:
This is honestly lovely. I would take a quick pass over the syntax tests and make sure that the widening cases are covered with the ops, but otherwise I'm very much in favor of this.

Issue #255, opened 2020-09-08
Add low-priority materializations for Handle in terms of ApplicativeError


djspiewak:
Given an `ApplicativeError[F, E]`, we should be able to get a `Handle[F, E]` (which implies `Raise[F, E]`). This would be very useful in that it allows us to get something like `Raise[IO, Throwable]`, which extends the generality of the library somewhat. It shouldn't cause any ambiguity issues if it is pushed to the lowest priority of the resolution chain in the `Handle` companion.

pchpsky:
@djspiewak I'd like to solve this issue. I need to add instance of `Handle` for `ApplicativeError` in `HandleInstances`. Is it correct?

djspiewak:
@pchpsky Almost! You need to create a `HandleLowPriorityInstances` which `HandleInstances` extends. This is what pushes the `ApplicativeError` materialization to a lower priority.

pchpsky:
@djspiewak thank you. I'll work on this.

PR #256, opened 2020-09-08
Get Handle from ApplicativeError

pchpsky:
Solves #255


djspiewak (changes requested):
Would you mind writing a test for this? I would create a `class` or something specifically within the test, then `implicit def` an `ApplicativeError` for that class (without implementing any of the methods; just define them all to be `???`) and try to get a `Handle` for that particular class. It would also be worth checking a few inductive materializations over that type (e.g. `Handle[Kleisli[Foo, Unit, *], ...]`) just to show that there's an unambiguous resolution.

Sorry this is a bit vague. Let me know if you want more specificity. :-)

djspiewak (approved):
This is perfect, thank you!

PR #262, opened 2020-10-02
add Local instances for State and function

enzief:


enzief:
I'm not sure who should be asked for review, so if you don't mind @djspiewak please.

LukaJCB (approved):
This is cool, thanks

kubukoz (approved):

kubukoz:
I'd consider having a compile-time test that ensures instances are resolved without extra hassle ;)

PR #279, opened 2020-11-16
move Local instances into Ask companion

enzief:
Follows https://github.com/typelevel/cats/pull/3043, moves typeclass instances from `Local` to `Ask` companion.


rossabaker:
I'm sorry this got buried.  Closing and reopening to see where we're at from a bincompat perspective...

rossabaker:
Yeah, this has binary compatibility ramifications.

I'm trying to think of the best way to test against a large corpus of MTL code if we get a MiMa-compliant solution.

PR #306, opened 2021-01-26
Add Tell instance for RWST

marcinzh:
Add missing instance of Tell for RWST


LukaJCB (approved):
Ah, thanks for fixing this! :)

PR #342, opened 2021-06-06
Create effect module w/ instances for IO

armanbilge:
So far I've added instances for `Stateful` and `Handle` for `IO`. Any others I can provide? Or for other datatypes in CE3?

### DIFF
--- a/tests/src/test/scala/cats/mtl/tests/IOTests.scala
+++ b/tests/src/test/scala/cats/mtl/tests/IOTests.scala
@@ -18,13 +18,14 @@ package cats.mtl.tests
 
 import cats.effect.testkit.TestInstances
 import cats.mtl.effect.instances.io._
-import cats.mtl.laws.discipline.{HandleTests, StatefulTests}
 import cats.effect.IO
 import cats.effect.IOLocal
 import munit.CatsEffectSuite
 import munit.DisciplineSuite
 
 class IOTests extends CatsEffectSuite with DisciplineSuite with TestInstances {
+  import cats.mtl.laws.discipline.{HandleTests, StatefulTests}
+
   implicit val ticker: Ticker = Ticker()
 
   // Terrible hack!
--- a/tests/src/test/scala/cats/mtl/tests/IOTests.scala
+++ b/tests/src/test/scala/cats/mtl/tests/IOTests.scala
@@ -21,17 +21,15 @@ import cats.mtl.effect.instances.io._
 import cats.mtl.laws.discipline.{HandleTests, StatefulTests}
 import cats.effect.IO
 import cats.effect.IOLocal
+import munit.CatsEffectSuite
+import munit.DisciplineSuite
 
-class IOTests extends BaseSuite with TestInstances {
-
+class IOTests extends CatsEffectSuite with DisciplineSuite with TestInstances {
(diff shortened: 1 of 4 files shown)


tests/src/test/scala/cats/mtl/tests/IOTests.scala
```diff
@@ -0,0 +1,37 @@
...
+
+class IOTests extends BaseSuite with TestInstances {
+
+  implicit val ticker: Ticker = Ticker()
+
+  IOLocal(0)
+    .flatMap { implicit local =>
+      IO.delay(checkAll("Stateful[IO]", StatefulTests[IO, Int].stateful))
+    }
+    .unsafeRunAndForget()
```
armanbilge, line 33: This test doesn't work right now, and IDK how to make it work. The tricky part is to create an instance of `Stateful[IO, Int]` we need an `IOLocal[Int]` first which only comes in an `IO`.

djspiewak: https://github.com/typelevel/munit-cats-effect

Have a look at the `Resource` support! :-)

armanbilge: Thanks!! Looking

ChristopherDavenport:
So I think this is a great idea, but for development this does build a cycle

cats-effect depends on cats-mtl for Test scope. 

So whenever a new binary incompatible scala version drops this will need to not be released. Then cats-effect released, and then this module released.

I still think this is probably the right place. But its probably worth noting.

armanbilge:
> So I think this is a great idea, but for development this does build a cycle

Oops. Would this be much easier as a CE module then?

tests/src/test/scala/cats/mtl/tests/IOTests.scala
armanbilge: This is now hacked/fixed 🙃

armanbilge:
Via Discord: @djspiewak suggests adding this module would be "dodgy". Alternatives are that this goes in Cats Effect or I spin up a microlib. Also pinging @oleg-py for his thoughts.

Also want to reference #120 which (possibly in some variation) could also find a happy home with effect-related stuff.

Finally, can someone help me figure out the build failure? I don't understand what it's referencing. Thanks!

tests/src/test/scala/cats/mtl/tests/IOTests.scala
```diff
@@ -18,13 +18,14 @@ package cats.mtl.tests
...
 import cats.mtl.effect.instances.io._
-import cats.mtl.laws.discipline.{HandleTests, StatefulTests}
 import cats.effect.IO
 import cats.effect.IOLocal
 import munit.CatsEffectSuite
 import munit.DisciplineSuite
 
 class IOTests extends CatsEffectSuite with DisciplineSuite with TestInstances {
+  import cats.mtl.laws.discipline.{HandleTests, StatefulTests}
+
```
armanbilge, line 28: I don't understand why the imports need to be here, but the other tests do it too.

armanbilge:
I ended up putting these odds and ends together in a microlib: https://github.com/armanbilge/oxidized

Issue #440, opened 2022-07-25
Scaladoc link is invalid now


floating-cat:
Two places:

1. In [homepage](https://typelevel.org/cats-mtl/index.html)'s right-top bar, click the doc link would go to https://www.javadoc.io/doc/org.typelevel/cats-mtl-docs_2.13/1.3.0/ and it will jump to the `javadoc.io`'s homepage with "Artifact org.typelevel:cats-mtl-docs_2.13 does not exist." error.
The right link seems to be https://www.javadoc.io/doc/org.typelevel/cats-mtl_2.13/.

2. In this repo's [markdown file](https://github.com/typelevel/cats-mtl#documentation), it uses the https://typelevel.org/cats-mtl/api/ for Scaladoc which would jump to the https://www.javadoc.io/doc/org.typelevel/cats-mtl-docs_2.13/1.3.0/ too.

benhutchison:
I think the `unidocs` sbt project which is supposed to upload the `cats-mtl-docs` artifact isnt getting built.

I wonder if this advice is relevant
`// Make sure to add to your root aggregate so it gets published!`
https://typelevel.org/sbt-typelevel/site.html#how-can-i-publish-unidoc-api-docs

tyukiand:
As of today (2023-03-21), both the link on the homepage and in the README.md are (again?) giving the

> Artifact org.typelevel:cats-mtl-docs_2.13 does not exist. 

error message on javadoc.io.

Could [this idea with the version bump](https://github.com/typelevel/cats-mtl/pull/459#issuecomment-1423566433) still be relevant?

PR #459, opened 2022-11-11
Add unidocs to root aggregate

b-vennes:
The `cats-mtl-docs` artifact generated by the `unidocs` module is not being published because the `unidocs` module is not part of the root aggregate.

The `cats-mtl-docs` unified documentation should now be published and viewable with javadoc.io, which will fix the broken links discovered by #440.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,7 +29,7 @@ lazy val commonNativeSettings = Seq(
 
 val CatsVersion = "2.8.0"
 
-lazy val root = tlCrossRootProject.aggregate(core, laws, tests)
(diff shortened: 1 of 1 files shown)


armanbilge (approved):
Thanks, I think this was my mistake!

We will need to publish a new cats-mtl for this to take affect.

rossabaker:
Thanks!  I'm not sure why there's no CI.  Closing and reopening to see if that gives it a nudge...

rossabaker:
We have enough other things on main that the next one is going to have to be a minor release.  If we can't bring that in for a landing real fast, I'll cherry-pick this and make a 1.3.1 just to fix the site.

PR #465, opened 2022-12-28
implement Cats typeclass instances for Ask and Local

bpholt:
This lets users `map`/`flatMap` (for `Ask`) or `imap` (for `Local`) the environment of a given instance and make the mapped value available in a new implicit scope. 

This came up for me when trying to provide a `Local[F, natchez.Span[F]]`, where `F[_]` is `Kleisli[IO, Span[IO], *]`. `Local.baseLocalForKleisli` gives us `Local[Kleisli[IO, Span[IO], *], Span[IO]]`, which is not quite the right shape for an app with an `implicit ev: Local[F, Span[F]]` (which becomes `ev: Local[Kleisli[IO, Span[IO], *], Span[Kleisli[IO, Span[IO], *]]]`). In other words, `baseLocalForKleisli` gives us a local environment of `Span[IO]` when we need `Span[Kleisli[IO, Span[IO], *]]`.

Given `Invariant[Local[F, *]]`, we can do 
```scala
Local[Kleisli[IO, Span[IO], *], Span[IO]].imap(Kleisli.liftK)(Kleisli.applyK(NoopSpan()))
```
to get the `Local[Kleisli[IO, Span[IO], *], Span[Kleisli[IO, Span[IO], *]]]` needed for the app.

### DIFF
--- a/tests/shared/src/test/scala/cats/mtl/tests/LocalCatsInstanceTests.scala
+++ b/tests/shared/src/test/scala/cats/mtl/tests/LocalCatsInstanceTests.scala
@@ -21,7 +21,7 @@ package tests
 import cats.data._
 import cats.laws.discipline.InvariantTests
 import cats.laws.discipline.eq._
-import org.scalacheck.Arbitrary
+import org.scalacheck.{Arbitrary, Gen}
 import org.scalacheck.Arbitrary._
 
 class LocalCatsInstanceTests extends BaseSuite {
@@ -29,17 +29,9 @@ class LocalCatsInstanceTests extends BaseSuite {
       implicit ev: Eq[F[B]]): Eq[Kleisli[F, A, B]] =
     Eq.by((x: Kleisli[F, A, B]) => x.run)
 
-  private implicit def arbLocalReader[F[_]: Monad, AA: Arbitrary]
-      : Arbitrary[Local[Kleisli[F, AA, *], AA]] = Arbitrary {
-    arbitrary[AA].map { a =>
-      new Local[Kleisli[F, AA, *], AA] {
-        override def local[A](fa: Kleisli[F, AA, A])(f: AA => AA): Kleisli[F, AA, A] =
-          fa.local(f)
-        override def applicative: Applicative[Kleisli[F, AA, *]] = implicitly
-        override def ask[E2 >: AA]: Kleisli[F, AA, E2] = Kleisli.pure(a)
-      }
-    }
-  }
+  private implicit def arbLocalReader[F[_]: Monad, AA]
+      : Arbitrary[Local[Kleisli[F, AA, *], AA]] =
+    Arbitrary(Gen.const(Local.baseLocalForKleisli))
 
   private implicit def eqLocal[F[_], A](implicit E: Eq[F[A]]): Eq[Local[F, A]] = Eq.by(_.ask)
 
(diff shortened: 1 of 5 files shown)


rossabaker:
Interesting.  I was thinking about your `Eq[Ask]`, and it hit me: is there anywhere in the Cats ecosystem where we have an instance for an instance?  I can see where this is useful, but it's novel enough to have me thinking why it hasn't done.

bpholt:
I have a vague memory of suggesting similar sorts of instances in the past, but I'm having trouble finding those PRs. Maybe it was in chat, not sure. I don't remember there being specific reasons not to do it, mostly just "huh that's different". But it's been a while; maybe I'm misremembering.

I considered just adding the `map`/`flatMap`/`imap` methods to `Ask` and `Local` directly, but after using cats-tagless for a while, it's been my experience that it's occasionally frustrating to have types with the methods for Cats instances, that don't actually implement the cats instances. Implementing the functionality using the typeclasses felt like it would be a little more flexible, and all else being equal, that's a good thing.

Implementing this way also lets us take advantage of the laws, which would be harder to test otherwise. Now that I think about it, that actually might be the stronger argument for doing it this way. 🙂

armanbilge:
> where we have an instance for an instance

Do the `Invariant[Eq]` instances and stuff count? They're not higher-kinded, but they are still instances right?

https://github.com/typelevel/cats/blob/024dd3d13ac9725ea663a4f914a89aa41bc3baea/core/src/main/scala/cats/Invariant.scala#L171-L255

tests/shared/src/test/scala/cats/mtl/tests/LocalCatsInstanceTests.scala
```diff
@@ -0,0 +1,50 @@
...
+    Eq.by((x: Kleisli[F, A, B]) => x.run)
+
+  private implicit def arbLocalReader[F[_]: Monad, AA: Arbitrary]
+      : Arbitrary[Local[Kleisli[F, AA, *], AA]] = Arbitrary {
+    arbitrary[AA].map { a =>
+      new Local[Kleisli[F, AA, *], AA] {
+        override def local[A](fa: Kleisli[F, AA, A])(f: AA => AA): Kleisli[F, AA, A] =
+          fa.local(f)
+        override def applicative: Applicative[Kleisli[F, AA, *]] = implicitly
+        override def ask[E2 >: AA]: Kleisli[F, AA, E2] = Kleisli.pure(a)
```
armanbilge, line 39: Sorry, I'm a bit confused about this instance. Shouldn't this be `Kleisli.ask` ?
```suggestion
        override def ask[E2 >: AA]: Kleisli[F, AA, E2] = Kleisli.ask
```

armanbilge (commented):
👍 I really like this idea!

tests/shared/src/test/scala/cats/mtl/tests/LocalCatsInstanceTests.scala
bpholt: It's slightly different, because `Kleisli.ask` won't use the arbitrary value generated a couple lines above, but the tests pass both ways and I'm not sure now why I wrote it that way before. I simplified things in ebcbf14 (basically accepting your suggestion).

rossabaker (approved):

armanbilge (approved):

PR #476, opened 2023-02-09
Be less precise about Scala 3 version

rossabaker:
We're already built for Scala 3.2 (perhaps since @benhutchison's PR).  I think it's fair to just say "Scala 3": most Scala 3 projects know to get on the latest by this point.


armanbilge (approved):
Agree 👍

rossabaker:
Sorry, Arman, disregard the re-request.

PR #478, opened 2023-02-09
Change abstract instance vals to defs

### LINKED ISSUE #264: Inconsistent use of def vs val in embedded typeclass
In `Ask`:
```scala
def applicative: Applicative[F]
```
In `Censor`:
```scala
val applicative: Applicative[F]
```

So if you define traits that extend both, like:
```scala
trait MyTrait[F[_], Env, Evt]  extends Censor[F, Chain[Evt]] with Ask[F, Env]
trait MyTrait2[F[_], Env, Evt] extends Ask[F, Env] with Censor[F, Chain[Evt]]
```
The order becomes important. `MyTrait2` compiles but `MyTrait` fails with:
```
stable, immutable value required to override:
[error] val applicative: cats.Applicative[F] (defined in trait Censor)
[error]   with def applicative: cats.Applicative[F] (defined in trait Ask)
[error]   trait MyTrait[F[_], Env, Evt] extends Censor[F, Chain[Evt]] with Ask[F, Env]
```

How about using the same pattern to avoid such issues?

rossabaker:
Fixes #264.  I standardized on def: there are more of those, and MiMa seems okay with it.


build.sbt
```diff
@@ -35,7 +35,9 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .crossType(CrossType.Pure)
   .settings(name := "cats-mtl")
   .settings(
-    libraryDependencies += "org.typelevel" %%% "cats-core" % CatsVersion
+    libraryDependencies += "org.typelevel" %%% "cats-core" % CatsVersion,
+    libraryDependencies += "org.typelevel" %%% "cats-effect" % "3.4.5",
+
```
armanbilge, line 40: Adding CE would make a circular dependency. Also, why? 🤔

rossabaker: Because this was my initial playground experimenting with IOLocal and I didn't clean up after myself.

[hangs head in shame]

armanbilge (approved):
`def`s are more flexible for implementors as well: you can always implement a `def` with a `val`, but not the other way around.

rossabaker:
I'll wait a bit for an official maintainer since this one is more consequential than a README change.  (Specifically: I don't think we can go back if Arman and I are deemed wrong.)

PR #514, opened 2023-08-03
Add `Raise#fromEither` and friends

armanbilge:
Inspired by discussion with @benhutchison in https://github.com/typelevel/cats-effect/discussions/3765#discussioncomment-6623490. The idea is to add a bunch of methods for lifting from the "primitive" monads `Either`, `Reader`, `Writer`, `State` into `F[_]`. This complements https://github.com/typelevel/cats-mtl/pull/489 which pushes towards using `IO` for `F[_]` instead of the monad transformers `EitherT`, `Kleisli`, `WriterT`, and `StateT`. So we need a mechanism for lifting.

- `Ask#fromReader`
- `Raise#fromEither` (analogous to `ApplicativeError#fromEither`)
- `Stateful#fromState`
- `Tell#fromWriter`

### DIFF
--- a/core/src/main/scala/cats/mtl/Ask.scala
+++ b/core/src/main/scala/cats/mtl/Ask.scala
@@ -53,6 +53,9 @@ trait Ask[F[_], +E] extends Serializable {
   def reader[A](f: E => A): F[A] = applicative.map(ask)(f)
 
   def fromReader[A](ra: Reader[E, A]): F[A] = reader(ra.run)
+
+  def fromKleisli[A](ka: Kleisli[F, E, A])(implicit F: FlatMap[F]): F[A] =
+    ask.flatMap(ka.run(_))
 }
 
 private[mtl] trait AskForMonadPartialOrder[F[_], G[_], E] extends Ask[G, E] {
--- a/core/src/main/scala/cats/mtl/Ask.scala
+++ b/core/src/main/scala/cats/mtl/Ask.scala
@@ -17,7 +17,7 @@
 package cats
 package mtl
 
-import cats.data.{Kleisli, ReaderWriterStateT => RWST}
+import cats.data.{Kleisli, Reader, ReaderWriterStateT => RWST}
 import cats.mtl.Ask.{const, AskImpl}
 import cats.syntax.all._
 
@@ -51,6 +51,8 @@ trait Ask[F[_], +E] extends Serializable {
   def ask[E2 >: E]: F[E2]
 
   def reader[A](f: E => A): F[A] = applicative.map(ask)(f)
+
+  def fromReader[A](ra: Reader[E, A]): F[A] = reader(ra.run)
 }
 
 private[mtl] trait AskForMonadPartialOrder[F[_], G[_], E] extends Ask[G, E] {
--- a/core/src/main/scala/cats/mtl/Raise.scala
+++ b/core/src/main/scala/cats/mtl/Raise.scala
@@ -79,6 +79,10 @@ trait Raise[F[_], -E] extends Serializable {
   def ensure[E2 <: E, A](fa: F[A])(error: => E2)(predicate: A => Boolean)(
       implicit A: Monad[F]): F[A] =
     A.flatMap(fa)(a => if (predicate(a)) A.pure(a) else raise(error))
+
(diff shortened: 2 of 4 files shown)


core/src/main/scala/cats/mtl/Ask.scala
```diff
@@ -51,6 +51,11 @@ trait Ask[F[_], +E] extends Serializable {
   def ask[E2 >: E]: F[E2]
 
   def reader[A](f: E => A): F[A] = applicative.map(ask)(f)
+
+  def fromReader[A](ra: Reader[E, A]): F[A] = reader(ra.run)
+
+  def fromKleisli[A](ka: Kleisli[F, E, A])(implicit F: FlatMap[F]): F[A] =
+    ask.flatMap(ka.run(_))
```
armanbilge, line 58: Andrew raised a great question about use-cases, which made me realize we can do even better 🙂 

For example suppose you have a `Ask[IO, Request[IO]]` and you want to interop with an http4s thing `Kleisli[IO, Request[IO], Response[IO]]`. Well, now you can!

armanbilge:
> As one user of this library, I'd like to "bias towards action" and see you merge this @armanbilge

@benhutchison I literally cannot merge my own PRs :) I can however review and merge work submitted by others.

core/src/main/scala/cats/mtl/Raise.scala
```diff
@@ -79,6 +79,13 @@ trait Raise[F[_], -E] extends Serializable {
   def ensure[E2 <: E, A](fa: F[A])(error: => E2)(predicate: A => Boolean)(
       implicit A: Monad[F]): F[A] =
     A.flatMap(fa)(a => if (predicate(a)) A.pure(a) else raise(error))
+
+  def fromEither[A](ea: Either[E, A])(implicit F: Applicative[F]): F[A] =
+    ea.fold(raise, F.pure)
+
+  def fromEitherT[E2 <: E, A](ea: EitherT[F, E2, A])(implicit F: Monad[F]): F[A] =
+    F.flatMap(ea.value)(fromEither(_))
+
```
igor-ramazanov, line 88: Also useful

```suggestion

  def fromOption[E2 <: E, A](oa: Option[A])(e: => E2)(implicit F: Applicative[F]): F[A] =
    oa.fold[F[A]](raise(e))(F.pure)

  def fromOptionT[E2 <: E, A](ota: OptionT[F, A])(e: => E2)(implicit F: Monad[F]): F[A] =
    F.flatMap(ota.value)(oa => fromOption(oa)(e))

```

core/src/main/scala/cats/mtl/Raise.scala
```diff
@@ -79,6 +79,19 @@ trait Raise[F[_], -E] extends Serializable {
...
+  def fromOption[E2 <: E, A](oa: Option[A])(e: => E2)(implicit F: Applicative[F]): F[A] =
+    oa.fold[F[A]](raise(e))(F.pure)
+
+  def fromOptionT[E2 <: E, A](ota: OptionT[F, A])(e: => E2)(implicit F: Monad[F]): F[A] =
+    F.flatMap(ota.value)(oa => fromOption(oa)(e))
+
```
igor-ramazanov, line 94:
```suggestion
  def fromValidated[E2 <: E, A](va: Validated[E2, A])(implicit F: Applicative[F]): F[A] =
    va.fold(raise, F.pure)

```

igor-ramazanov: I wonder if we are not accidentally adding a performance penalty comparing with, for example:
```scala
  def fromValidated[E2 <: E, A](va: Validated[E2, A])(implicit F: Applicative[F]): F[A] =
    va match {
      case Validated.Invalid(e) => raise(e)
      case Validated.Valid(a) => F.pure(a)
    }

```

armanbilge: Actually we are. Using `fold` requires allocating lambdas, pattern matching does not. I don't think it's important in practice.

djspiewak (approved):
Bike-shedding a bit, the `from` prefix threw me for quite a while as I was trying to parse this. I can't think of a better name though. This is absolutely a nice addition.

PR #559, opened 2024-04-08
Change version to 1.4.0 in the documentation

ChidiRnweke:
The release of the library is 1.4.0 instead of 1.3.0 as per https://mvnrepository.com/artifact/org.typelevel/cats-mtl


rossabaker (approved):
Thanks!

Issue #562, opened 2024-04-17
Support Scala Native 0.5


mpilquist:
Depends on https://github.com/typelevel/cats/issues/4587.

PR #577, opened 2024-08-04
Update Scala Native to 0.5.4

lolgab:
Fixes https://github.com/typelevel/cats-mtl/issues/562


rossabaker (approved):
I suppose this jump will require a minor bump, similar to Cats.

PR #592, opened 2024-11-19
Use `cats.syntax.all._` instead of `cats.implicits`

lenguyenthanh:
similar to https://github.com/typelevel/cats/pull/4394


sh0hei (approved):
Thanks, @lenguyenthanh !

PR #619, opened 2025-03-06
 Implemented submarine error propagation for `Handle`

lenguyenthanh:
This is a follow up of #489 on behalf of @djspiewak, the description below is copied from previous pr with some minor update of the examples.

This adds builder syntax to Handle which makes it possible to materialize an instance of Handle for any error type given an `ApplicativeThrow[F]`. Syntax is intentionally reminiscent of `try/catch`:

### Scala 2 example

```scala
sealed trait DomainError
object DomainError {
  case object Failed extends DomainError
  case object Derped extends DomainError
}

def foo(implicit h: Raise[F, DomainError]): F[String] = ???
def bar(implicit h: Handle[F, DomainError]): F[String] = ???

Handle.allowF[F, DomainError] { implicit h =>
  foo *> bar
} rescue {
  case DomainError.Failed => ???
  case DomainError.Derped => ???
}
```

### scala 3 example

```scala
enum DomainError:
  case Failed
  case Derped

def foo(using h: Raise[F, DomainError]): F[String] = ???
def bar(using h: Handle[F, DomainError]): F[String] = ???

Handle.allow:
  foo *> bar
.rescue:
  case DomainError.Failed => ???
  case DomainError.Derped => ???
```

You can see some other examples in the tests. You can suspend as many different error types as you want and it all composes the way you would expect.

This effectively eliminates the need for `EitherT` or `IorT` outside of tests (where both can still be useful for building ad-hoc harnesses that are less powerful than `IO`). Types infer quite nicely on both Scala 2 and Scala 3, though obviously Scala 3's `using` syntax makes this even nicer.


core/src/main/scala-3/cats/mtl/HandleVariant.scala
```diff
@@ -0,0 +1,47 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats
+package mtl
+
+trait HandleVariant { this: Handle.type =>
```
armanbilge, line 20: This is more of the convention:
```suggestion
private[mtl] trait HandleCrossCompat { this: Handle.type =>
```

core/src/main/scala/cats/mtl/Handle.scala
```diff
@@ -217,6 +218,41 @@ private[mtl] trait HandleInstances extends HandleLowPriorityInstances {
...
 
-object Handle extends HandleInstances {
+object Handle extends HandleInstances with HandleVariant {
+
   def apply[F[_], E](implicit ev: Handle[F, E]): Handle[F, E] = ev
+
+  def allowF[F[_], E]: AdHocSyntaxTired[F, E] =
+    new AdHocSyntaxTired[F, E]
+
+  final class AdHocSyntaxTired[F[_], E] {
```
armanbilge, line 228: We can do tricks to avoid the allocations.
```suggestion
  final class AdHocSyntaxTired[F[_], E](private val unit: Unit) extends AnyVal {
```

core/src/main/scala/cats/mtl/Handle.scala
```diff
@@ -217,6 +218,41 @@ private[mtl] trait HandleInstances extends HandleLowPriorityInstances {
...
+
+    def apply[A](body: Handle[F, E] => F[A])(implicit F: ApplicativeThrow[F]): Inner[A] =
+      new Inner(body)
+
+    final class Inner[A](body: Handle[F, E] => F[A])(implicit F: ApplicativeThrow[F]) {
+      def rescue(h: E => F[A]): F[A] = {
```
armanbilge, line 234: Another trick to avoid allocations, although this slightly changes the ergonomics:
```suggestion
    final class Inner[A](private val body: Handle[F, E] => F[A]) extends AnyVal {
      def rescue(h: E => F[A])(implicit F: ApplicativeThrow[F]): F[A] = {
```

core/src/main/scala-3/cats/mtl/HandleVariant.scala
```diff
@@ -0,0 +1,47 @@
...
+  import Handle.Submarine
+
+  def allow[E]: AdHocSyntaxWired[E] =
+    new AdHocSyntaxWired[E]
+
+  final class AdHocSyntaxWired[E]:
```
armanbilge, line 26: Since this is Scala 3, I think with some clever use of `inline` we could make all the allocations / boilerplate disappear from the generated code.

lenguyenthanh: Another fancy way to do it is using polymorphic function:
```diff
-  inline def allow[E]: AdHocSyntaxWired[E] =
-    new AdHocSyntaxWired[E]()
-
-  private[mtl] final class AdHocSyntaxWired[E]:
-    inline def apply[F[_], A](inline body: Handle[F, E] ?=> F[A]): InnerWired[F, E, A] =
-      new InnerWired(body)
+  def allow[E]: [F[_], A] => (Handle[F, E] ?=> F[A]) => InnerWired[F, E, A] =
+    [F[_], A] => (body: Handle[F, E] ?=> F[A]) => InnerWired(body)
```
Generated code is a bit nicer compare to `inline` solution:

```java
public final class mtl$minussubmarine$package$
implements Serializable {
    public static final mtl$minussubmarine$package$ MODULE$ = new mtl$minussubmarine$package$();

    private mtl$minussubmarine$package$() {
    }

    private Object writeReplace() {
        return new ModuleSerializationProxy(mtl$minussubmarine$package$.class);
    }

    public EitherT<Eval, Throwable, String> test1() {

        return (EitherT)((HandleCrossCompat.InnerWired)Handle$.MODULE$.allow().apply((Function1 & Serializable)contextual$1 -> {
            Error$ error$ = (Error$)all$.MODULE$.toRaiseOps((Object)Error$.MODULE$);
            return (EitherT)package.all$.MODULE$.toFunctorOps(RaiseOps$.MODULE$.raise$extension((Object)error$, (Raise)contextual$1), (Functor)EitherT$.MODULE$.catsDataMonadErrorForEitherT((Monad)Eval$.MODULE$.catsBimonadForEval())).as((Object)"nope");
        })).rescue((Function1 & Serializable)x$1 -> {
            Error$ error$ = x$1;
            if (Error$.MODULE$.equals(error$)) {
                String string = (String)package.all$.MODULE$.catsSyntaxApplicativeId((Object)"error");
                return (EitherT)ApplicativeIdOps$.MODULE$.pure$extension((Object)string, (Applicative)EitherT$.MODULE$.catsDataMonadErrorForEitherT((Monad)Eval$.MODULE$.catsBimonadForEval()));
            }
            throw new MatchError((Object)error$);
        }, (ApplicativeError)EitherT$.MODULE$.catsDataMonadErrorForEitherT((Monad)Eval$.MODULE$.catsBimonadForEval()));
    }
}
```

core/src/main/scala/cats/mtl/Handle.scala
```diff
@@ -217,6 +218,42 @@ private[mtl] trait HandleInstances extends HandleLowPriorityInstances {
...
-object Handle extends HandleInstances {
+object Handle extends HandleInstances with HandleCrossCompat {
+
   def apply[F[_], E](implicit ev: Handle[F, E]): Handle[F, E] = ev
+
+  def allowF[F[_], E]: AdHocSyntaxTired[F, E] =
+    new AdHocSyntaxTired[F, E](())
+
+  @scala.annotation.nowarn("msg=dubious usage of method hashCode with unit value")
+  private[mtl] final class AdHocSyntaxTired[F[_], E](private val unit: Unit) extends AnyVal {
```
armanbilge, line 229: Huh, that's a weird warning 😅 instead of `Unit` we could also use `dummy: Byte` or something else.

core/src/main/scala-3/cats/mtl/HandleVariant.scala
armanbilge: Nice! It looks like we are still allocating `InnerWired`, do you think we can make that one go away too?

lenguyenthanh: > Nice! It looks like we are still allocating `InnerWired`, do you think we can make that one go away too?

I tried but couldn't find a way. `opaque` doesn't work because `opaque` doesn't work with context function.

I tried with value class by extending `AnyVal` because of `Value classes may not be a member of another class` error.

Maybe there is another way? Let me try some more.

btw, which approach do you prefer polymorphic function or inlining?

armanbilge: > btw, which approach do you prefer polymorphic function or inlining?

I think the polymorphic function is allocating something? to call the `apply` on?
```java
Handle$.MODULE$.allow().apply
```

lenguyenthanh: I think you're right, let me investigate.

lenguyenthanh: hit a compiler crash while trying to make `InnerWired` extends `AnyVal`: https://github.com/scala/scala3/issues/22752

core/src/main/scala/cats/mtl/Handle.scala
lenguyenthanh: fixed in https://github.com/typelevel/cats-mtl/pull/619/commits/a70b9f4e5d9f0ca9c50b4f932ca5387aaa7777e9

core/src/main/scala-3/cats/mtl/HandleVariant.scala
lenguyenthanh: > > btw, which approach do you prefer polymorphic function or inlining?
> 
> I think the polymorphic function is allocating something? to call the `apply` on?
> 
> ```java
> Handle$.MODULE$.allow().apply
> ```

yes it is:

```java
    default public <E> Function1 allow() {
        return new Function1(this){
            private final /* synthetic */ HandleCrossCompat $outer;
            {
                if ($outer == null) {
                    throw new NullPointerException();
                }
                this.$outer = $outer;
            }

            public Object apply(Object body) {
                return new InnerWired<F, E, A>(this.$outer, arg_0 -> HandleCrossCompat.cats$mtl$HandleCrossCompat$$anon$1$$_$apply$$anonfun$1(body, arg_0));
            }

            private static /* synthetic */ Object $deserializeLambda$(SerializedLambda serializedLambda) {
                return LambdaDeserialize.bootstrap("lambdaDeserialize", new MethodHandle[]{cats$mtl$HandleCrossCompat$$anon$1$$_$apply$$anonfun$1(java.lang.Object cats.mtl.Handle )}, serializedLambda);
            }
        };
    }
```

lenguyenthanh: tried extension function to replace `InnerWired` like this:

```scala
private[mtl] trait HandleCrossCompat { this: Handle.type =>

  inline def allow[E]: AdHocSyntaxWired[E] =
    new AdHocSyntaxWired[E]()

  private[mtl] final class AdHocSyntaxWired[E]:

    inline def apply[F[_], A](inline body: Handle[F, E] ?=> F[A]): HF[F, E, A] =
      body
}

type HF[F[_], E, A] = Handle[F, E] ?=> F[A]
object HF:
  import Handle.Submarine
  extension [F[_], E, A](body: HF[F, E, A])
    def rescue(h: E => F[A]): ApplicativeThrow[F] ?=> F[A] =
      val Marker = new AnyRef

      def inner[B](fb: F[B])(f: E => F[B]): F[B] =
        ApplicativeThrow[F].handleErrorWith(fb):
          case Submarine(e, Marker) => f(e.asInstanceOf[E])
          case t => ApplicativeThrow[F].raiseError(t)

      given Handle[F, E] with
        def applicative = Applicative[F]
        def raise[E2 <: E, B](e: E2): F[B] =
          ApplicativeThrow[F].raiseError(Submarine(e, Marker))
        def handleWith[B](fb: F[B])(f: E => F[B]): F[B] = inner(fb)(f)

      inner(body)(h)
```
but this also failed miserably, because `allow` will eagerly try to consume `Handle[F,E]`, so we can't never reach `rescue`.

core/src/main/scala-3/cats/mtl/HandleCrossCompat.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  inline def allow[E]: AdHocSyntaxWired[E] =
+    new AdHocSyntaxWired[E]()
+
+  private[mtl] final class AdHocSyntaxWired[E]:
+    inline def apply[F[_], A](inline body: Handle[F, E] ?=> F[A]): InnerWired[F, E, A] =
+      new InnerWired(body)
+
+  private[mtl] final class InnerWired[F[_], E, A](body: Handle[F, E] ?=> F[A]):
+    def rescue(h: E => F[A]): ApplicativeThrow[F] ?=> F[A] =
```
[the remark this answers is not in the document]
lenguyenthanh: yes, it does, context function required exactly one extra allocation.It's amazed, that you know these little details.

core/src/main/scala-3/cats/mtl/HandleVariant.scala
lenguyenthanh: @armanbilge do you have any idea to avoid `InnerWired`, I'm out of idea 😓

djspiewak (approved):
At the risk of playing "obama giving obama a medal" here, I'm going to click the approve button. :D I haven't been keeping up too closely on this (thank you for taking over @lenguyenthanh!!); are there any outstanding items that would keep us from merging and releasing this?

lenguyenthanh:
Thanks @djspiewak, you deserved a medal for this 🏅 

> are there any outstanding items that would keep us from merging and releasing this?

There are some small issues, I don't think they're blocker but I'll list them for clarity: 

- I'd love to add some docs for this new feature (can be in a separated pr but should be done before releasing)
- @armanbilge suggested that there can be another layer of indirection that can be removed (but I couldn't find it tho 😄 ).
- error message can be a surprise for users (and leaking internal implementation details), for example:

```scala

type F[A] = EitherT[Eval, Throwable, A]

def test: F[String] =
  allow[Error.type]:
    Error.raise[F, String].as("nope")
//   .rescue:
//     case Error => "error".pure[F]


/*
[error] ./mtl-submarine.scala:21:3
[error] Found:    cats.mtl.InnerWired[[A] =>> cats.data.EitherT[cats.Eval, Throwable, A],
[error]   Error.type, String]
[error] Required: F[String]
[error] 
*/
```

The last one is annoyed me the most, but it seems we can't improve upon this :(

lenguyenthanh:
added some preliminary docs, so I'm happy if we can merge this!

cc @armanbilge as you also put a lot of though in this one.

lenguyenthanh:
example of using this technique in a wild: https://github.com/lichess-org/lila-search/pull/542

djspiewak (approved):

djspiewak (changes requested):
Oh actually quick trivial question: can you update `build.sbt` to bump the base version to 1.6?

lenguyenthanh:
> Wow that is very satisfyingly slick…

yes, it soooooo is!


> can you update `build.sbt` to bump the base version to 1.6?

done

djspiewak (approved):

PR #625, opened 2025-03-18
Add `LiftValue` and `LiftKind`

NthPortal:
Add `LiftValue` and `LiftKind`.

Make `MonadPartialOrder` a subtype of `LiftValue`.

The motivation for this PR is the desire to add a `LocalLogger` to log4cats that can store log context in a `cats.mtl.Local`. log4cats loggers have a `mapK` method; supporting `mapK` on a `LocalLogger` requires being able to transform a `Local[F, *]` into a `Local[G, *]`, which in turn requires the functionality exposed by ~~`KindTransformer`~~ **`LiftKind`**.

`LiftKind` (formerly `KindTransformer`) is originally (and currently) from otel4s, designed by me (so we don't need to worry about attributing it to otel4s as I am simply personally (re)licensing my own code/design to typelevel/cats-mtl here).

### DIFF
--- a/core/src/main/scala/cats/mtl/LiftValue.scala
+++ b/core/src/main/scala/cats/mtl/LiftValue.scala
@@ -1,0 +1,211 @@
+/*
+ * Copyright 2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats
+package mtl
+
+import cats.data.{ReaderWriterStateT => RWST, _}
+
+import scala.annotation.implicitNotFound
+import scala.{unchecked => uc}
+
+/** Lifts values from the higher-kinded type `F` to the higher-kinded type `G`. */
+@implicitNotFound("no way defined to lift values from ${F} to ${G}")
+trait LiftValue[F[_], G[_]] extends (F ~> G) {
+
+  /** An [[cats.Applicative `Applicative`]] instance for the higher-kinded type `F`. */
+  def applicativeF: Applicative[F]
+
+  /** An [[cats.Applicative `Applicative`]] instance for the higher-kinded type `G`. */
+  def applicativeG: Applicative[G]
+
+  /**
+   * Lifts `F[A]` to `G[A]`.
+   *
+   * @note
+   *   This method is usually best implemented by a `liftF` method on `G`'s companion object.
+   */
+  def apply[A](fa: F[A]): G[A]
+
+  /** @return an instance that lifts `E` to `F` and then `F` to `G` */
+  def compose[E[_]](that: LiftValue[E, F]): LiftValue[E, G] =
+    LiftValue.Composed(that, this)
+
+  /** @return an instance that lifts `F` to `G` and then `G` to `H` */
+  def andThen[H[_]](that: LiftValue[G, H]): LiftValue[F, H] =
+    LiftValue.Composed(this, that)
+}
+
+object LiftValue extends LowPriorityLiftValueInstances {
+
+  private[mtl] class Identity[F[_]](implicit val applicative: Applicative[F])
+      extends LiftValue[F, F] {
+    final def applicativeF: Applicative[F] = applicative
+    final def applicativeG: Applicative[F] = applicative
+    final def apply[A](fa: F[A]): F[A] = fa
+  }
+
+  private[this] final class Composed[F[_], G[_], H[_]] private (
+      inner: LiftValue[F, G],
+      outer: LiftValue[G, H]
+  ) extends LiftValue[F, H] {
+    def applicativeF: Applicative[F] = inner.applicativeF
+    def applicativeG: Applicative[H] = outer.applicativeG
+    def apply[A](fa: F[A]): H[A] = outer(inner(fa))
+  }
+
+  private object Composed {
+    def apply[F[_], G[_], H[_]](
+        inner: LiftValue[F, G],
+        outer: LiftValue[G, H]
+    ): LiftValue[F, H] =
+      if (inner.isInstanceOf[Identity[F @uc]]) outer.asInstanceOf[LiftValue[F, H]]
+      else if (outer.isInstanceOf[Identity[G @uc]]) inner.asInstanceOf[LiftValue[F, H]]
+      else new Composed(inner, outer)
+  }
+
+  def apply[F[_], G[_]](implicit lv: LiftValue[F, G]): LiftValue[F, G] = lv
+
+  implicit def id[F[_]: Applicative]: LiftValue[F, F] = new Identity[F]
+
+  /*
+   The compositional implicits are sufficient to derive all implicit instances,
+   but non-compositional ones with higher priority are provided to avoid
+   unnecessary allocation.
+   */
+
+  implicit def eitherT[F[_], L](implicit F: Monad[F]): LiftValue[F, EitherT[F, L, *]] =
+    eitherTImpl(F)
+
+  implicit def iorT[F[_], L: Semigroup](implicit F: Monad[F]): LiftValue[F, IorT[F, L, *]] =
+    iorTImpl(F)
+
+  implicit def kleisli[F[_], A](implicit F: Applicative[F]): LiftValue[F, Kleisli[F, A, *]] =
+    kleisliImpl(F)
+
+  implicit def optionT[F[_]](implicit F: Monad[F]): LiftValue[F, OptionT[F, *]] =
+    optionTImpl(F)
+
+  implicit def rwst[F[_], E, L: Monoid, S](
+      implicit F: Monad[F]
+  ): LiftValue[F, RWST[F, E, L, S, *]] =
+    rwstImpl(F)
+
+  implicit def stateT[F[_], S](implicit F: Monad[F]): LiftValue[F, StateT[F, S, *]] =
+    stateTImpl(F)
+
+  implicit def writerT[F[_], L: Monoid](
+      implicit F: Applicative[F]
+  ): LiftValue[F, WriterT[F, L, *]] =
+    writerTImpl(F)
+}
+
+private[mtl] sealed trait LowPriorityLiftValueInstances {
+  protected[this] def eitherTImpl[F[_], L](F: Monad[F]): LiftValue[F, EitherT[F, L, *]] =
+    new LiftValue[F, EitherT[F, L, *]] {
+      implicit val applicativeF: Monad[F] = F
+      val applicativeG: Applicative[EitherT[F, L, *]] = EitherT.catsDataMonadErrorForEitherT
+      def apply[A](fa: F[A]): EitherT[F, L, A] = EitherT.liftF(fa)
+    }
+
+  implicit def eitherTComposed[F[_], G[_], L](
+      implicit inner: LiftValue[F, G],
+      G: Monad[G]
+  ): LiftValue[F, EitherT[G, L, *]] =
+    inner.andThen(eitherTImpl(G))
+
+  protected[this] def iorTImpl[F[_], L: Semigroup](F: Monad[F]): LiftValue[F, IorT[F, L, *]] =
+    new LiftValue[F, IorT[F, L, *]] {
+      implicit val applicativeF: Monad[F] = F
+      val applicativeG: Applicative[IorT[F, L, *]] = IorT.catsDataMonadErrorForIorT
+      def apply[A](fa: F[A]): IorT[F, L, A] = IorT.liftF(fa)
+    }
+
+  implicit def iorTComposed[F[_], G[_], L: Semigroup](
+      implicit inner: LiftValue[F, G],
+      G: Monad[G]
+  ): LiftValue[F, IorT[G, L, *]] =
+    inner.andThen(iorTImpl(G))
+
+  protected[this] def kleisliImpl[F[_], A](F: Applicative[F]): LiftValue[F, Kleisli[F, A, *]] =
+    new LiftValue[F, Kleisli[F, A, *]] {
+      implicit val applicativeF: Applicative[F] = F
+      val applicativeG: Applicative[Kleisli[F, A, *]] = Kleisli.catsDataApplicativeForKleisli
+      def apply[B](fa: F[B]): Kleisli[F, A, B] = Kleisli.liftF(fa)
+    }
+
+  implicit def kleisliComposed[F[_], G[_], A](
+      implicit inner: LiftValue[F, G]
+  ): LiftValue[F, Kleisli[G, A, *]] =
+    inner.andThen(kleisliImpl(inner.applicativeG))
+
+  protected[this] def optionTImpl[F[_]](F: Monad[F]): LiftValue[F, OptionT[F, *]] =
+    new LiftValue[F, OptionT[F, *]] {
+      implicit val applicativeF: Monad[F] = F
+      val applicativeG: Applicative[OptionT[F, *]] = OptionT.catsDataMonadForOptionT
+      def apply[A](fa: F[A]): OptionT[F, A] = OptionT.liftF(fa)
+    }
+
+  implicit def optionTComposed[F[_], G[_]](
+      implicit inner: LiftValue[F, G],
+      G: Monad[G]
+  ): LiftValue[F, OptionT[G, *]] =
+    inner.andThen(optionTImpl(G))
+
+  protected[this] def rwstImpl[F[_], E, L: Monoid, S](
+      F: Monad[F]
+  ): LiftValue[F, RWST[F, E, L, S, *]] =
+    new LiftValue[F, RWST[F, E, L, S, *]] {
+      implicit val applicativeF: Monad[F] = F
+      val applicativeG: Applicative[RWST[F, E, L, S, *]] =
+        IndexedReaderWriterStateT.catsDataMonadForRWST
+      def apply[A](fa: F[A]): RWST[F, E, L, S, A] = RWST.liftF(fa)
+    }
+
+  implicit def rwstComposed[F[_], G[_], E, L: Monoid, S](
+      implicit inner: LiftValue[F, G],
+      G: Monad[G]
+  ): LiftValue[F, RWST[G, E, L, S, *]] =
+    inner.andThen(rwstImpl(G))
+
+  protected[this] def stateTImpl[F[_], S](F: Monad[F]): LiftValue[F, StateT[F, S, *]] =
+    new LiftValue[F, StateT[F, S, *]] {
+      implicit val applicativeF: Monad[F] = F
+      val applicativeG: Applicative[StateT[F, S, *]] =
+        IndexedStateT.catsDataMonadForIndexedStateT
+      def apply[A](fa: F[A]): StateT[F, S, A] = StateT.liftF(fa)
+    }
+
+  implicit def stateTComposed[F[_], G[_], S](
+      implicit inner: LiftValue[F, G],
+      G: Monad[G]
+  ): LiftValue[F, StateT[G, S, *]] =
+    inner.andThen(stateTImpl(G))
+
+  protected[this] def writerTImpl[F[_], L: Monoid](
+      F: Applicative[F]
+  ): LiftValue[F, WriterT[F, L, *]] =
+    new LiftValue[F, WriterT[F, L, *]] {
+      implicit val applicativeF: Applicative[F] = F
+      val applicativeG: Applicative[WriterT[F, L, *]] = WriterT.catsDataApplicativeForWriterT
+      def apply[A](fa: F[A]): WriterT[F, L, A] = WriterT.liftF(fa)
+    }
+
+  implicit def writerTComposed[F[_], G[_], L: Monoid](
+      implicit inner: LiftValue[F, G]
+  ): LiftValue[F, WriterT[G, L, *]] =
+    inner.andThen(writerTImpl(inner.applicativeG))
+}
--- a/laws/src/main/scala/cats/mtl/laws/LiftValueLaws.scala
+++ b/laws/src/main/scala/cats/mtl/laws/LiftValueLaws.scala
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats
+package mtl
+package laws
+
+import cats.laws.IsEq
+import cats.laws.IsEqArrow
+
+trait LiftValueLaws[F[_], G[_]] {
+  implicit def lift: LiftValue[F, G]
+  implicit final def F: Applicative[F] = lift.applicativeF
+  implicit final def G: Applicative[G] = lift.applicativeG
+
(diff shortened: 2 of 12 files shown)


NthPortal:
I don't know how to write laws/tests for either of these commits, and would greatly appreciate assistance doing so

NthPortal:
@iRevive if/when this lands, we should replace `org.typelevel.otel4s.KindTransformer` in otel4s with a deprecated type alias that points at `cats.mtl.KindTransformer`.

also I will need to remember to make a PR to cats-effect adding an implicit method to create `KindTransformer` instances for `Resource` to `Resource`'s companion

armanbilge:
How does this relate to `MonadPartialOrder`?
https://github.com/typelevel/cats-mtl/blob/ef624597f7c29178e237dcd9f09101c3997e1a90/core/src/main/scala/cats/mtl/MonadPartialOrder.scala#L29

NthPortal:
@armanbilge good question. `MonadPartialOrder` is certainly insufficient in that `Monad` is not a sufficient context bound in all cases (`cats.effect.Resource` for example requires `MonadCancel`), and is also over-constraining for all but one of implicits provided in the companion (only `StateT` requires `Monad`; most of the rest only require `Functor` and `Kleisli` and identity have no context bounds at all). *if* `Monad` was sufficient in general, then simply adding `limitedMapK` and `liftFunctionK` to `MonadPartialOrder` would be a perfectly fine alternative to adding `KindTransformer`, but that is not the case.

... does that answer your question?

NthPortal:
cc @rossabaker do you recall us discussing something related to Arman's question? I have a vague recollection of it, though it's been quite a long time

armanbilge:
Yes, thanks, that's a very helpful answer! 

> and is also over-constraining for all but one of the implicits provided in the companion (only `StateT` requires `Monad`; most of the rest only require `Functor` and `Kleisli` and identity have no context bounds at all)

Can you explain more about this? Specifically this part:

> `Kleisli` and identity have no context bounds at all

To clarify, do you mean _lifting_ into `Kleisli` has no context bounds? Because to have a `Monad[Kleisli[F, E, *]]` you do need a `Monad[F]`.

core/src/main/scala/cats/mtl/KindTransformer.scala
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+  implicit def iorT[F[_]: Functor, L]: KindTransformer[F, IorT[F, L, *]] =
+    new KindTransformer[F, IorT[F, L, *]] {
+      val liftK: F ~> IorT[F, L, *] = IorT.liftK
+      def limitedMapK[R](ga: IorT[F, L, R])(f: F ~> F): IorT[F, L, R] =
+        ga.mapK(f)
+    }
+
+  implicit def kleisli[F[_], A]: KindTransformer[F, Kleisli[F, A, *]] =
```
NthPortal, line 97: @armanbilge this. the implicit method that returns/provides `KindTransformer[F, Kleisli[F, A, *]]` does not require a `Monad[F]` or even a `Functor[F]`

armanbilge: Aha, right. Thanks for clarifying!

core/src/main/scala/cats/mtl/KindTransformer.scala
```diff
@@ -0,0 +1,110 @@
...
+   * `G`.
+   *
+   * @note
+   *   Implementors SHOULD NOT override this method; the only reason it is not final is for
+   *   optimization of the identity case.
+   */
+  def liftFunctionK(f: F ~> F): G ~> G =
+    new (G ~> G) {
+      def apply[A](ga: G[A]): G[A] = limitedMapK(ga)(f)
+    }
```
NthPortal, line 65: for simplicity, you can think of `limitedMapK` as a specialisation/optimisation of `liftFunctionK`; the purpose of `KindTransformer` is to provide both `F ~> G` and `(F ~> F) => (G ~> G)`

iRevive:
Eventually we can use the KindTransformer from cats-mtl in otel4s, right?

NthPortal:
@iRevive we may need to think about making an otel4s release with all the changes that don't require cats-effect 3.6 in order to adopt this change in a timely manner, and if we don't, we may end up causing users to deal with conflicts if/when this ends up in use (e.g. in log4cats)

rossabaker:
This abstraction has certainly proven itself useful in otel4s, and again in this new case.  Because cats-mtl is on the classpath _everywhere_ now, it's important we get this right.  I'm trying to scavenge for prior art.

* @milessabin shared shapeless-3's [`FunctorK`](https://github.com/typelevel/shapeless-3/blob/main/modules/deriving/src/test/scala/shapeless3/deriving/type-classes.scala#L327-L328), also in [cats-tagless](https://typelevel.org/cats-tagless/typeclasses.html#functorK).  An [earlier draft](https://github.com/typelevel/otel4s/pull/284#issuecomment-1652538888) is closer, but still not the same.
* Skunk has an [`Exchange`](https://github.com/typelevel/skunk/blob/e522905cc7f7fef5b253cafee6bec11179ff3d73/modules/core/shared/src/main/scala/net/protocol/Exchange.scala), which is `F ~> F`, which is a higher-kinded `cats.Endo`.  But that type is just an argument to `limitedMapK`, not the full abstraction.  Neither `Exchange` nor a hypothetical `EndoK` turn up much.
* I have been unsuccessful in Hoogling for prior art in Haskell.

I wonder, since this is more general than monads, does it belong in mtl?  Is it perhaps cats?  Without laws, it would be banished to alleycats, which would not work for the places we want to use this.

NthPortal:
> I wonder, since this is more general than monads, does it belong in mtl? Is it perhaps cats?

I have no objection to it being in cats "proper".

> Without laws, it would be banished to alleycats

I have no doubt that laws are writable, I just don't personally know how, and am also not the most versed in category theory to know what mathematical/logical/categorical laws there are.

> shapeless-3's `FunctorK`

If we were trying to `FunctorK` for `LocalLogger[F[_]]` (for example), would we be creating `FunctorK[LocalLogger]`? how would one derive that? I can't think of how to do it without the functionality provided by `KindTransformer`

rossabaker:
This is allowed in our current lawlessness:

```scala
  def silly[F[_], G[_]](f: F ~> G): KindTransformer[F, G] =
    new KindTransformer[F, G] {
      val liftK: F ~> G = f
      def limitedMapK[A](ga: G[A])(f: F ~> F): G[A] = ga
    }
```

How about this to relate the operations?

```
limitedMapK(liftK(fa))(f) <-> liftK(f(fa))
```

NthPortal:
we should also relate `limitedMapK` and `liftFunctionK` based on the latter's default implementation (thus asserting that a given implementation is equivalent to the default):

```scala
limitedMapK(ga)(f) <-> liftFunctionK(f)(ga)
```

is there anything concrete we can say about individual methods? for example, we could test that the identity function does nothing, as follows:

```scala
limitedMapK(ga)(FunctionK.id) <-> ga
```

but idk if that's useful or meaningful. I don't think anything can be said of `liftK` without relating it to the other methods as it's essentially an arbitrary function

NthPortal:
@rossabaker what about laws for testing `mapK` on `Ask` and `Local`?

laws/src/main/scala/cats/mtl/laws/KindTransformerLaws.scala
```diff
@@ -0,0 +1,47 @@
...
+
+trait KindTransformerLaws[F[_], G[_]] {
+  implicit def ktInstance: KindTransformer[F, G]
+
+  // external law:
+  def limitedMapKIdentityIsPure[A](ga: G[A]): IsEq[G[A]] =
+    ktInstance.limitedMapK(ga)(FunctionK.id) <-> ga
+
+  // internal laws:
+  def liftKThenLimitedMapKIsMapThenLiftK[A](fa: F[A], f: F ~> F): IsEq[G[A]] =
```
NthPortal, line 33: I kind of hate this law name

laws/src/main/scala/cats/mtl/laws/discipline/KindTransformerTests.scala
```diff
@@ -0,0 +1,55 @@
...
+  implicit val ktInstance: KindTransformer[F, G]
+
+  def laws: KindTransformerLaws[F, G] = KindTransformerLaws[F, G]
+
+  def kindTransformer[A](
+      implicit arbFA: Arbitrary[F[A]],
+      arbGA: Arbitrary[G[A]],
+      arbFF: Arbitrary[F ~> F],
+      eqGA: Eq[G[A]]
+  ): RuleSet =
```
NthPortal, line 37: I think  `List` is a good candidate for `F` here (as it yields many simple `F ~> F` implementations such as `drop(n)`, `take(n)`)

rossabaker:
> I have no objection to it being in cats "proper".

Some historical points why `FunctorK` didn't make the cut.  I don't know how many of those apply here.

> I have no doubt that laws are writable, I just don't personally know how, and am also not the most versed in category theory to know what mathematical/logical/categorical laws there are.

I'm just winging it too.  This isn't an abstraction handed down on high by the abstract algebrists, this is one that scratches an itch.  (Some are like that!)
 
> If we were trying to `FunctorK` for `LocalLogger[F[_]]` (for example), would we be creating `FunctorK[LocalLogger]`? how would one derive that? I can't think of how to do it without the functionality provided by `KindTransformer`

Yeah, I think `FunctorK` abstracts over things with a `mapK`, but it isn't enough to define one.  For that, we need something like this, that unites liftK and mapK.

> we should also relate limitedMapK and liftFunctionK based on the latter's default implementation

Agreed.

> `limitedMapK(ga)(FunctionK.id) <-> ga`

I started thinking along these lines, and then I stopped because I thought it was implied by the other one, but I don't think it's quite the same.  This seems like a good one.

> what about laws for testing mapK on Ask and Local?

That's a fair question.  Our existing `mapK` tests are geared toward data types, not for options that exist on a typeclass that hitherto didn't have `mapK`.  If `FunctorK` were deemed useful, maybe it should sit beneath `Ask`!

laws/src/main/scala/cats/mtl/laws/discipline/KindTransformerTests.scala
rossabaker: Do we have to write that arbitrary ourself, or does Scalacheck provide one out of the box, like it does for Function (with the help of Cogen).

rossabaker:
I suppose that depends on the definition of "good".

- It abstracts over enough to be useful to derive instances of multiple type classes.
- `Applicative[F]` is enough for all use cases in this library, and `Applicative[G]` for all but `Stateful`.  Maybe that would be interesting for something like `Nested`, but I don't think it has ever come up.
- It's lawless, so mischievous implementations only reveal themselves when testing derived instances:
    ```scala
  implicit def monadPartialOrderForOptionT[F[_]](
      implicit F: Monad[F]): MonadPartialOrder[F, OptionT[F, *]] =
    new MonadPartialOrder[F, OptionT[F, *]] {
      val monadF = F
      val monadG = OptionT.catsDataMonadForOptionT[F]
      def apply[A](fa: F[A]) = OptionT.none /* yikes */
    }
    ```

    Put another way, nothing currently enforces what's in the header comment about "the lifting distributes".

NthPortal:
perhaps a different question would be better: why is it important that a single typeclass be able to derive things, and not for example having methods require both a (theoretical) `LiftK` and `Applicative`?

NthPortal:
the other option is we make this a microlib instead

rossabaker:
An implicit `F ~> G` doesn't make a lot of sense to me unless it expresses a relationship to other operations (for `MonadPartialOrder`, the `Monad`; for `KindTransformer`, the `limitedMapK`).

A microlib would lower the stakes of getting it right in cats-mtl, unless cats-mtl depends on that microlib anyway.  And otel4s will only continue to get more important, and needs this functionality.

rossabaker:
The closest prior art I've found to `KindTransformer` Haskell is [Inject](https://hackage.haskell.org/package/functor-combinators-0.4.1.3/docs/Data-Functor-Combinator.html#t:Inject).  The `hmap` is like `mapK`.  The `inject` is like `liftK`, except the `F` remains in the signature.  Frustratingly, it doesn't seem to enforce any lawful relationship between the two as we've tried to do here!

NthPortal:
> unless cats-mtl depends on that microlib anyway

it would not; just otel4s (and log4cats) would, probably

NthPortal:
@rossabaker I think we've been thinking about this all wrong, and should consider it from an entirely different angle.

the purpose of `KindTransformer` is to enable the transformation of an arbitrary type parameterised by `F[_]` into being parameterised by `G[_]`, without requiring any other value or object that converts between `F` and `G` (though it does not aim to be the only context bound—a `MonadCancelThrow[G]` for example might still be required). for this reason, I'm no longer confident that it necessarily belongs in cats-mtl.

we should consider what transformations `KindTransformer` currently enables, and what types might need additional capabilities to be transformed from being parameterised by `F` to `G`. `#liftK` enables transforming methods with no inputs in `F` but an output in `F` into having their output in `G`. `limitedMapK`/`liftFunctionK` enables transforming methods with a single input in `F` of an arbitrary type `A` and an output of `F[A]` into having a single input and output of `G[A]` instead (i.e. scoping/context change).

what types cannot be transformed by the current capabilities of `KindTransformer`? is there an existing type, or can we at least contrive a type?

NthPortal:
I think a better description/name for `limitedMapK` and `liftFunctionK` is "lift scope"

NthPortal:
btw @rossabaker, having laws that assert consistency does not prevent you from writing degenerate instances. for example:

```scala
new KindTransformer[F, OptionT[F, *]] {
  val liftK: F ~> OptionT[F, *] = new (F ~> OptionT[F, *]) {
    def apply[A](fa: F[A]): OptionT[F, A] = OptionT.none
  }
  def limitedMapK[A](ga: OptionT[F, A])(f: F ~> F): OptionT[F, A] = OptionT.none
}
```

laws/src/main/scala/cats/mtl/laws/KindTransformerLaws.scala
```diff
@@ -0,0 +1,47 @@
...
+    ktInstance.limitedMapK(ktInstance.liftK(fa))(f) <-> ktInstance.liftK(f(fa))
```
NthPortal, line 34: this isn't true for `StateT`, for reasons I do not fully understand

NthPortal:
I've done some work on refining the abstraction(s) [here](https://github.com/NthPortal/lifts/pull/1)

laws/src/main/scala/cats/mtl/laws/discipline/KindTransformerTests.scala
NthPortal: we have/had to write it ourselves

NthPortal:
@djspiewak I'm going to organise this reply by addressing shorter and/or simpler points first.

for continuity, what was previously called `KindTransformer` is now called `LiftKind`. `LiftValue` is a subset of `LiftKind`'s capabilities and a generalisation of the functionality of `LiftIO`.

I did a thorough exploration of possible typeclass hierarchies over at NthPortal/lifts#1.

--------

> `Kleisli`, which is really inconveniently defined by `R => F[A]`. Except it's not, and it's actually `F[R => F[A]]`

I think you must be mixing it up with something else, as `Kleisli` has precisely the first definition you give.

--------

> I do want to be clear that I'm not saying that `KindTransformer` is _definitely unsound_!

`LiftKind` has a number of laws that I think show its soundness, and they are now in this PR as well as my exploration.

--------

> The `(F ~> F) => (G ~> G)` thing does make me a little nervous, since it implies you have the contravariant transformation (`G ~> F`) somewhere in your back pocket

neither `LiftKind` nor `LiftValue` provide a transformation `G ~> F`, and I don't personally see the implication. to me, `LiftValue` only implies `G ~> F` specifically for the immediate result of lifting a value, in that it must be guaranteed to be unliftable to be lawful. that guarantee is the basis of `Unlift` and some of the laws.

--------

> I've name-dropped `FunctorK`

`FunctorK` doesn't work well as an abstraction as it assumes (incorrectly!) that the `mapK` operation has no context bounds. `Resource` from cats-effect is a counterexample in wide use for which a typeclass with `mapK` functionality would be very helpful.

`MapK` and `LiftScope` in my exploration expose the same (or a limited version of that) capability while still supporting context bounds.

--------

> You're trying to take multiple types of effects (such as as maybe optionality and an error channel) and smash them together such that you can have computations which have access to all of those capabilities within a uniform side-channel.

> A classic "obvious" solution to effect composition materializes from the `Traverse` typeclass, which seems to give a way of defining a pretty general flavor of `flatMap` which cuts across multiple monads simultaneously.

`LiftKind` actually does not attempt to do what you describe, and I don't think comparing it to `Traverse` is useful or meaningful. if you want access to the capabilities of any particular effect that was composed using `LiftKind`, you will need to manually decompose it. `LiftValue` allows you to "wrap" or "box" a type, and `LiftKind` extends those capabilities to keep track of how to perform certain transformations (specifically scoping operations) on the underlying "wrapped" or "boxed" type.

one of the key use cases of `LiftKind` is to be able to use telemetry and observability types uniformly across `HttpApp[F]` and `HttpRoutes[F]` from http4s, which de-alias to `Http[F, F]` and `Http[OptionT[F, *], F]` respectively. we'd like to be able to perform tracing and logging in `HttpRoutes` even though it doesn't necessarily have behaviour or a response for every possible request—hence the `OptionT` when it is de-aliased. `LiftKind` allows us to to lift a tracer into `OptionT` and still perform scoping operations with it (such as creating spans). if `F` happens to be an `EitherT`, it doesn't give us access to that in any way.

--------

> MPO is really neat because it basically gives us a way to _sort of_ talk about how `OptionT[IO, A]` contains an `IO` without having to figure out how to talk about decomposing the latter into the former (since as noted, this isn't possible in general).

for reference, this is the same as the functionality `LiftValue` has, except `LiftValue` doesn't require two `Monad` instances.

> and a _few_ laws

there may be laws that apply to it, but there are no laws/tests for `MonadPartialOrder` in cats-mtl.

> This gives us a _little_ bit of power (which we use in Cats MTL to reduce boilerplate)

while `MonadPartialOrder` is sufficient to derive instances for (for example) `Ask`, it is not *necessary*, which leads me my more general point: I do not like `MonadPartialOrder`.

I hate `MonadPartialOrder`'s name. 'partial order' is already a term with a meaning—and a typeclass in cats no less—and `MonadPartialOrder` does not relate to that meaning. to the extent that it *might* relate to the meaning of 'partial order', it only has two monads, and a set of two elements is either fully ordered or entirely unordered (and it would be vacuous to call it partially ordered). but the relationship between the two monads isn't even one of ordering, but one of derivability—you can derive one from the other.

much more significantly than the name though, I think `MonadPartialOrder` is a bad abstraction, and I will again use `Ask` as an example. `MonadPartialOrder[F, G]` has 3 parts: `Monad[F]`, `Monad[G]`, and `F ~> G`. for comparison, `LiftValue` is just `F ~> G`. in order to derive `Ask[G, ?]` from `Ask[F, ?]`, you need `F ~> G` and `Applicative[G]` (arguably `Functor[G]`, but that's #635). `MonadPartialOrder[F, G]` requires an extraneous `Monad[F]` and `FlatMap[G]`(ish) that the `Ask[G, ?]` doesn't need. even if you argue that `Monad[F]` is implicit in the existence of `Monad[G]`, there's still unnecessary `FlatMap` capability. I opine that only the bare necessities for deriving `Ask[G, ?]` from `Ask[F, ?]` should be required—`LiftValue[F, G]` and `Applicative[G]`.

NthPortal:
oops, I missed a point I had intended to address.

> My _fear_ is that this is secretly why the laws don't work for `StateT`

the reason `StateT` is not lawful is because `StateT[F, S, A]` is `F[S => F[(S, A)]]`; in other words, because of the double wrapping in `F`. this makes `StateT#mapK` unlawful for use in any abstraction over the operation `mapK`

NthPortal:
one way or another, some version of this code will continue to exist either in otel4s or some dependency of otel4s. I (and probably others) think it belongs here, so I would like to reach a consensus. there are definite downsides to making it a microlib, though that's always an option as well

NthPortal:
@djspiewak does the fact that `Unlift` is part of the laws module and only exists for law testing make it more palatable?

djspiewak:
It still makes me pretty uncomfortable. IMO, the right way to leverage laws is to think of them as denotational rewrite rules. Similar to how you could define decrement in a Peano system in terms of rewrites: `forall n . dec (succ n) <-> n`. Put another way, in an ideal world (not the one we live in), there would be no (principled) code which can be written in terms of the typeclasses which cannot be rewritten according to the laws into progressively more canonical expression forms, ultimately boiling down to the axiomatic assumptions of our handwave calculus.

Obviously that's not really the world we live in, but it's kind of what I *ascribe* to. So that in turn means that we should try to avoid this type of law-specific machinery for the same reason that we try to avoid laws that are of the form `somethingComplex <-> true` (which Cats Effect 2 had a lot of).

Looking at the laws in question, I think you could accomplish a very similar set of denotational implications by adding laws which relate lifting to `FunctionK.id`. Between that and bare parametricity, I think you imply the relevant properties (which is also why this is how the applicative laws are structured).

NthPortal:
@djspiewak I confess, I've only been working with cats and pure-fp for a couple years, and am largely unfamiliar with category theory, so a lot of your comments have gone over my head.

I'm struggling to come up with any tests that use `FunctionK.id` other than the one that's already there. additionally, `MonadPartialOrder` does not currently have laws. what laws would you say it should have?

I'd also like to note that while `LiftValue`/`liftF` might be shaped like `pure`, I don't think it's quite the same; in particular, I can't think of a reason we should forbid instances like `LiftValue[Resource[F, *], Stream[F, *]]` (implemented by `fs2.Stream.resource`, of course), `LiftValue[Option, List]`, or `LiftValue[List, Vector]`, which are most definitely not analogous to `Applicative.pure`

djspiewak (changes requested):
I'd really like to see `Unlift` removed, but aside from that, LGTM

laws/src/main/scala/cats/mtl/laws/LiftValueLaws.scala
```diff
@@ -0,0 +1,57 @@
...
+  // external laws:
+  def liftFPureIsPure[A](value: A): IsEq[To[A]] =
+    liftInstance.liftF(From.pure(value)) <-> To.pure(value)
+
+  def liftFTupleIsTupleLiftF[A, B](fa: From[A], fb: From[B]): IsEq[To[(A, B)]] =
+    liftInstance.liftF(From.tuple2(fa, fb)) <->
+      To.tuple2(liftInstance.liftF(fa), liftInstance.liftF(fb))
+
+  def liftFAddsNoSideEffects[A, B](a: A, toB: To[B]): IsEq[To[B]] =
+    liftInstance.liftF(From.pure(a)) *> toB <-> toB
```
NthPortal, line 39: @djspiewak I've added these laws, though I'm not sure how good they are. they also seem to imply that `LiftValue` ought to contain `Applicative` instances for both of its types

NthPortal: how about this for `ap`?

```scala
  def liftFApIsApLiftF[A, B](fa: From[A], ff: From[A => B]): IsEq[To[B]] =
    lift.liftF(From.ap(ff)(fa)) <-> To.ap(lift.liftF(ff))(lift.liftF(fa))
```

core/src/main/scala/cats/mtl/LiftValue.scala
```diff
@@ -0,0 +1,162 @@
...
+   * @return
+   *   a higher-kinded function that lifts `From` into `To`.
+   *
+   * @note
+   *   This method is usually best implemented by a `liftK` method on `To`'s companion object.
+   */
+  def liftK: From ~> To =
+    new (From ~> To) {
+      def apply[A](fa: From[A]): To[A] = liftF(fa)
+    }
```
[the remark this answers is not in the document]
NthPortal: going with consistency with `MonadPartialOrder`, to reduce confusion

NthPortal:
> How does this relate to `MonadPartialOrder`?

@armanbilge after much design work, `LiftValue` is added with two subtypes: `MonadPartialOrder` and `LiftKind`

djspiewak (approved):
👏🏼

NthPortal:
(pushed some test coverage improvements (most of the compositional implicits weren't being tested), as well as a couple code comments explaining why there are seemingly unnecessary implicit methods)

NthPortal:
> as a general rule please don't force push to pull requests

I'll try to remember they for the future. I'm used to scala/scala's policy, which is... precisely the opposite lol. they prefer clean history

djspiewak:
> I'll try to remember they for the future. I'm used to scala/scala's policy, which is... precisely the opposite lol. they prefer clean history

Yeah it's quite a common policy! We're definitely in the minority (though, in an appeal to authority, Linux follows a similar policy to what we do). *It's the children who are wrong…*

PR #639, opened 2025-08-14
Add `Ask#liftTo` and `Local#liftTo`

NthPortal:


NthPortal:
I have a commit that adds these as implicit derivations rather than just manual ones, but it causes this line (https://github.com/typelevel/cats-mtl/blob/7d7441e6b4e9d404c930d37cf5247f1e9a8d5155/tests/shared/src/test/scala/cats/mtl/tests/SummonableImplicits.scala#L28) to fail with diverging implicit expansion, and I suspect that if you use a type that `Local` does not have a derivation for you will have the same issue with `Local`. or possibly I'm wrong, and we just need to add an explicit compositional derivation for `Kleisli` in `Ask` with higher priority than the `LiftValue` one and it will solve the problem. I don't know.

I don't actually care as much about having it be an implicit derivation, but I know others prefer that to having to explicitly call `liftTo`.

even if we get the implicits working properly, I would still find the explicit methods quite useful for when you have multiple `Ask`/`Local` instances next to each other (see https://github.com/typelevel/log4cats/pull/909/files#diff-e6d7b8e2b694bce277c5bdbee9a1f0ece134fe7049d15a460c6f8a73cdeb5039R114-R121 for example). in such cases, having to rely on implicit derivation would be incredibly painful and error-prone

djspiewak (approved):

PR #640, opened 2025-08-21
Mention Handle.allowF in Raise's implicitNotFound message

Prillan:
Example output:
```
scala> mayRaise[F]
               ^
       error: Could not find an implicit instance of Raise[F, Error]. If you have a
       good way of handling errors of type Error at this location, you may want
       to wrap the call-site in Handle.allowF, like so:

         Handle.allowF[F, Error] { implicit h =>
           // call code requiring Raise[F, Error]
         }.rescue { e: Error =>
           // handle the error
         }

       This is analogous to writing try/catch around this call ensuring that
       all errors of type Error are handled in the rescue block.

       Alternatively, you may want to construct a value of type EitherT for
       this call-site, rather than F.
       An example type:

         EitherT[F, Error, *]

       The EitherT will "catch" the errors of type Error.

       If you do not wish to handle errors of type Error at this location, you should
       add an implicit parameter of this type to your function. For example:

         (implicit fraise: Raise[F, Error])
``` 

Not sure what to do with Scala 3, is there an easier way to have a specific message to Scala 3 than splitting the trait into different files? Is it worth it?

### DIFF
--- a/core/src/main/scala/cats/mtl/Raise.scala
+++ b/core/src/main/scala/cats/mtl/Raise.scala
@@ -62,7 +62,7 @@ import scala.util.control.NonFatal
  * }}}
  */
 @implicitNotFound(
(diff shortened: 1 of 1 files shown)


djspiewak (approved):
This is great! It was on my bucket list but hadn't gotten to it. Thank you.

